PR #1, opened 2019-11-04
Create empty application

thewilkybarkid:
Sets up an empty Node container and corresponding Makefile. Covers both a production and development environment.


.travis/smoke-test.sh
```diff
@@ -0,0 +1,12 @@
...
+
+function finish {
+    make logs stop
+}
+
+trap finish EXIT
+
+make start
+
+.scripts/docker/wait-healthy.sh "${COMPOSE_PROJECT_NAME:-article-store}_app_1"
```
thewilkybarkid: Turns out `_1` isn't true.

.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: minimal
+
+services: docker
+
+env:
+  global:
+    - COMPOSE_PROJECT_NAME=articlestore
+
+install:
+  - ENV=dev make build
```
giorgiosironi: environment is problematic as this term, even if `NODE_ENV` is probably a standard here and there. These are targets or variants but there are many more environments in which an identical `prod` image could run.

.travis/smoke-test.sh
giorgiosironi: The `docker-compose` version on Travis CI had a bug related to this recently:
https://github.com/libero/jats-ingester/pull/56

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+
+#
+# Stage: Production environment
+#
+FROM node AS prod
+ENV NODE_ENV=production
+
+COPY --from=npm-prod /app/ .
+
+CMD ["sleep", "86400"]
```
giorgiosironi: does `sleep infinity` work in Alpine? Have seen this before

thewilkybarkid: Nope: https://github.com/libero/storybook/pull/2#discussion_r328218038

.travis/smoke-test.sh
thewilkybarkid: 7abe9f5ea3c108f83a70f2e7de2dd00d8185c636 avoids guessing.

README.md
```diff
@@ -1,6 +1,40 @@
 Libero Article Store
 ====================
 
+[![Build Status](https://travis-ci.com/libero/article-store.svg?branch=master)](https://travis-ci.com/libero/article-store)
```
thewilkybarkid, line 4: Split into https://github.com/libero/article-store/pull/2

docker-compose.override.yml
```diff
@@ -0,0 +1,6 @@
+version: '3.4'
+
+services:
+  app:
+    build:
+      target: dev
```
thewilkybarkid: Might be better to call the file `docker-compose.dev.yml` as it's abstract behind Make?

Makefile
```diff
@@ -0,0 +1,42 @@
...
+build: ## Build the containers
+	${DOCKER_COMPOSE} build
+
+start: ## Start the containers
+	${DOCKER_COMPOSE} up --detach
+
+stop: ## Stop the containers
+	${DOCKER_COMPOSE} down
+
+wait-healthy: ## Wait for the containers to be healthy
```
giorgiosironi, line 33: if no containers are running, this exits successfully:
```
$ make wait-healthy
for service in $(docker-compose $(if [[ $(if [[ ${ENV} = "prod" ]]; then echo "prod"; else echo "dev"; fi) = "prod" ]]; then echo "--file docker-compose.yml"; fi) ps --quiet); do \
        echo "${service}"; \
        .scripts/docker/wait-healthy.sh "${service}"; \
done
$ echo $?
0
```

docker-compose.override.yml
giorgiosironi: Eventually people will use custom `docker-compose` commands for debugging e.g. `docker-compose ps | grep ...` or `docker-compose logs ...`. As it stands this probably doesn't strictly need to be picked up there, but in general it should in case anything is added?

thewilkybarkid: Changes in 49c14b8a7f9f0e4a760e8a71cf8c3ea1278a7696. Docker Compose commands should go through Make, if it's just passing it through (So `make docker-compose "ps"` or something).

.travis.yml
giorgiosironi: the only standard I can find here is from `docker build`: `TARGET=dev`

thewilkybarkid: Hmm, not sure 'target' is the best, but let's go with it for now.

.travis.yml
```diff
@@ -1,4 +1,11 @@
...
+services: docker
+
+install:
+  - TARGET=prod make build
+  - TARGET=dev make build
+
 script:
-  - 'true'
+  - TARGET=prod .travis/smoke-test.sh
+  - TARGET=dev .travis/smoke-test.sh
```
thewilkybarkid, line 15: Turns out this is dangerous, as Docker Compose overrides the same built name, so both instances are the `dev` build.

giorgiosironi: The same image name? The `docker-compose.*.yml` files can force their own `image:` field which probably could have a `_dev` suffix. We did this before with a suffix on the tag?
https://github.com/libero/browser/blob/d4672d0ae647a2b01d14adb6e18e25c6d02d9977/docker-compose.test.yaml

Makefile
giorgiosironi: Add a `$DOCKER_COMPOSE ps --quiet | grep $COMPOSE_PROJECT_NAME` before which would fail if there are no matches?

giorgiosironi (approved):

Makefile
thewilkybarkid: Was thinking that this was a problem, but I guess if the container has already stopped it won't be listed.

thewilkybarkid: A bit more complicated than thought, but come up with cd3abe303d0f243abe0e2c4de9f4d3a4dc63c2c4.

PR #2, opened 2019-11-05
Enable Travis CI

thewilkybarkid:
Sets up initial Travis CI configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,4 @@
+language: minimal
+
+script:
+  - 'true'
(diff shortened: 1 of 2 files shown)


thewilkybarkid:
Assuming we want to carry on with Travis rather than try GitHub actions?

giorgiosironi (approved):

giorgiosironi:
> Assuming we want to carry on with Travis rather than try GitHub actions?

At the moment, it's the standard choice; unless the pressure on performance becomes too great.

PR #3, opened 2019-11-06
Add empty HTTP server

thewilkybarkid:
Adds a HTTP server on port `8080` that responds with `200 OK` to every request.

### DIFF
--- a/.travis/smoke-test.sh
+++ b/.travis/smoke-test.sh
@@ -8,3 +8,6 @@ function finish() {
 trap finish EXIT
 
 make start wait-healthy
+
+ping=$(curl -sS 'http://localhost:8080/' 2>&1)
+[[ $ping == 'OK' ]]
--- a/.docker/docker-compose.dev.yml
+++ b/.docker/docker-compose.dev.yml
@@ -6,5 +6,7 @@ services:
       target: dev
     image: libero/article-store:${IMAGE_TAG:-master}-dev
     volumes:
+      - ../src/:/app/src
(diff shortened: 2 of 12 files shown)


.travis/smoke-test.sh
```diff
@@ -8,3 +8,6 @@ function finish() {
 trap finish EXIT
 
 make start wait-healthy
+
+ping=$(curl -sS 'http://localhost:8080/' 2>&1)
+[[ $ping == 'OK' ]]
```
giorgiosironi, line 13: is that the response body? Any value is good

thewilkybarkid: Yeah, Koa's default when nothing is set.

giorgiosironi: https://github.com/libero/article-store/pull/3/files#diff-f98485da1849ae631ba5c04e5cec07dd seems to say this is the response status?

thewilkybarkid: Koa sets the body based on the status code.

thewilkybarkid: Might be we should just test the code.

thewilkybarkid: Possibly even redundant as it waits for the containers to be healthy? 🤔

giorgiosironi: Reachable from the host is more than container being healthy as it tests also ports and how they are forwarded. Good as it is

giorgiosironi (approved):

PR #6, opened 2019-11-06
Add test suite

thewilkybarkid:
Add tests using Jest.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,5 +6,8 @@
     "rootDir": "src",
     "alwaysStrict": true,
     "esModuleInterop": true
-  }
+  },
+  "exclude": [
+    "test"
+  ]
 }
--- a/.docker/docker-compose.dev.yml
+++ b/.docker/docker-compose.dev.yml
@@ -6,7 +6,9 @@ services:
       target: dev
     image: libero/article-store:${IMAGE_TAG:-master}-dev
     volumes:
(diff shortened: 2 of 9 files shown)


tsconfig.json
```diff
@@ -6,5 +6,8 @@
     "rootDir": "src",
     "alwaysStrict": true,
     "esModuleInterop": true
-  }
+  },
+  "exclude": [
+    "test"
+  ]
```
thewilkybarkid: Surprised I have to include this. Maybe `rootDir` doesn't do what I think it does?

giorgiosironi (approved):
Infrastructurally, all makes sense

tsconfig.json
stephenwf: You may need to re-add `node_modules` here, I'm not entirely sure. I think `rootDir` is just where your source that will be emitted is, the project will be relative to the `tsconfig.json`. `tsc` will still try to type-check things not in `rootDir`.
> The "exclude" property defaults to excluding the node_modules, bower_components, jspm_packages and <outDir> directories when not specified.

thewilkybarkid: https://github.com/microsoft/TypeScript/issues/9858 is relevant.

thewilkybarkid: I read that as use `"include": ["src"]` instead.

thewilkybarkid: https://github.com/libero/article-store/pull/8

PR #7, opened 2019-11-07
Add request/response logging

thewilkybarkid:
Logs all requests and responses to stdout.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,8 +1,10 @@
 import Koa, { Context } from 'koa';
+import logger from 'koa-logger';
 import { constants as statusCodes } from 'http2';
 
 const app = new Koa();
 
+app.use(logger());
 app.use(async ({ response }: Context): Promise<void> => {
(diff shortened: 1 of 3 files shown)


thewilkybarkid:
Will probably want to exclude the healthcheck.

giorgiosironi (approved):

src/app.ts
```diff
@@ -1,8 +1,10 @@
 import Koa, { Context } from 'koa';
+import logger from 'koa-logger';
 import { constants as statusCodes } from 'http2';
 
 const app = new Koa();
 
+app.use(logger());
```
thewilkybarkid, line 7: Probably want a way to disable this for the test suite.

diversemix (approved):

src/app.ts
thewilkybarkid: https://github.com/libero/article-store/pull/7/commits/cec623c88f9ec5b39430616202be5a8de3b719c3 works (for now).

PR #8, opened 2019-11-07
Fix TypeScript file inclusion

thewilkybarkid:
Correctly tells the TypeScript compiler to only include the `src` directory.

Refs https://github.com/libero/article-store/pull/6#discussion_r343194021.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,8 +3,10 @@
     "target": "es2019",
     "module": "commonjs",
     "outDir": "build",
-    "rootDir": "src",
     "alwaysStrict": true,
     "esModuleInterop": true
-  }
+  },
+  "include": [
+    "src"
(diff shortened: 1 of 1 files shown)


stephenwf (approved):

diversemix (approved):
That should work - though I tend to put in 
```
"include": ["src/**/*.ts"]
```

thewilkybarkid:
Just did a quick test, it does only include TS files (eg a `.js` is ignored unless `allowJs` is set).

diversemix:
> Just did a quick test, it does only include TS files (eg a `.js` is ignored unless `allowJs` is set).

Thanks! ... and does it recurse down sub-directories too?

thewilkybarkid:
Yep!

PR #9, opened 2019-11-08
Add a router

thewilkybarkid:
Now only responds `200 OK` to requests to the root, and `404 Not Found` elsewhere.


package.json
```diff
@@ -16,6 +18,7 @@
     "eslint-config-airbnb-base": "^14.0.0",
     "eslint-plugin-import": "^2.18.2",
     "eslint-plugin-jest": "^23.0.2",
+    "indefinite": "^2.3.1",
```
thewilkybarkid: Didn't add `@types/indefinite` due to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40238.

test/router.test.ts
```diff
@@ -0,0 +1,12 @@
...
+import createRouter, { Routes } from '../src/router';
+
+describe('router', (): void => {
+  const router = createRouter();
+
+  Object.entries(Routes).forEach(([key, value]) => {
+    it(`has ${indefinite(key)} route`, (): void => {
+      expect(router.route(value)).toBeTruthy();
+    });
+  });
```
thewilkybarkid: Feels... bit dirty.

diversemix: Which bit? ... The loop around the `it` is pretty standard but yes looks a tad ugly compared to something like python. The check that is truthy is pretty vague - did you want to tighten it? for example check its a function?

giorgiosironi: Rephrase to `has a route named|for ${key}` and avoid a 156KB library?
In general for a unit test it's unexpected to use test code that is vastly more complex than the unit under test, it may break more often that what it is testing, as encountered in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40238

test/router.test.ts
```diff
@@ -0,0 +1,12 @@
+import indefinite from 'indefinite';
+import createRouter, { Routes } from '../src/router';
+
+describe('router', (): void => {
+  const router = createRouter();
+
+  Object.entries(Routes).forEach(([key, value]) => {
+    it(`has ${indefinite(key)} route`, (): void => {
```
giorgiosironi: the scope of these tests should alert of inconsistencies between the `enum` and the routes, so it's helpful to localize that problem

package.json
```diff
@@ -2,12 +2,14 @@
   "private": true,
   "dependencies": {
     "koa": "^2.11.0",
-    "koa-logger": "^3.2.1"
+    "koa-logger": "^3.2.1",
+    "koa-router": "^7.4.0"
```
thewilkybarkid: Looks like this is abandoned and https://github.com/koajs/router should be used instead.

src/router.ts
```diff
@@ -1,10 +1,7 @@
 import Router from '@koa/router';
-import { Context } from 'koa';
 import http from 'http2';
-
-export enum Routes {
-  'EntryPoint' = 'entry-point',
-}
+import { Context } from 'koa';
+import Routes from './routes';
```
thewilkybarkid: Had to extract the enum to avoid future circular dependencies.

giorgiosironi (approved):

PR #10, opened 2019-11-08
Add entry point

thewilkybarkid:
Makes the root of the API a `schema:EntryPoint` with a basic `schema:name`.

~Depends on #9.~


thewilkybarkid:
The CI smoke test needs updating.

test/routes/entry-point.test.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    expect(graph).toHaveLength(1);
+
+    const object = graph[0];
+
+    expect(object['@id']).toBe('http://example.com/path-to/entry-point');
+    expect(object['@type']).toContain('http://schema.org/EntryPoint');
+    expect(object).toHaveProperty(['http://schema.org/name']);
+
+    expect(next).toHaveBeenCalled();
```
davidcmoulton: Might be able to tighten this up a bit?`.toHaveBeenCalledTimes()` and / or `.toHaveBeenCalledWith()` etc.

test/routes/entry-point.test.ts
```diff
@@ -0,0 +1,29 @@
...
+describe('entry-point', (): void => {
+  it('should return the entry point', async (): Promise<void> => {
+    const context = createContext('entry-point');
+    const next = jest.fn();
+
+    await entryPoint(context.router)(context, next);
+
+    const { response } = context;
+
+    expect(response.status).toEqual(200);
```
davidcmoulton: I don't think it's wrong to use `toEqual` here, but it's usually used to test deep equality. For primitive values, `toBe` might be slightly better (as on [line 23](https://github.com/libero/article-store/pull/10/files#diff-2923edc37cc590e9eebebcf1fac9b18dR23)).

src/routes/entry-point.ts
```diff
@@ -0,0 +1,20 @@
+import Router from '@koa/router';
+import http from 'http2';
+import { Context, Middleware, Next } from 'koa';
+import { schema } from 'rdf-namespaces';
```
giorgiosironi, line 3: recently created project, seems healthy

src/routes/entry-point.ts
```diff
@@ -0,0 +1,20 @@
...
+import url from 'url';
+import Routes from './index';
+
+export default (router: Router): Middleware => (
+  async ({ request, response }: Context, next: Next): Promise<void> => {
+    response.status = http.constants.HTTP_STATUS_OK;
+    response.body = {
+      '@id': url.resolve(request.origin, router.url(Routes.EntryPoint, {})),
```
giorgiosironi, line 10: want to test `request.origin` in practice, but seems the most flexible solution to allow simple load balancing. To be tested whether it allows reverse proxies such as gateways or CDNs using `X-Forwarded-Host`.

src/routes/entry-point.ts
```diff
@@ -0,0 +1,20 @@
...
+      '@type': schema.EntryPoint,
+      [schema.name]: { '@value': 'Article Store', '@language': 'en' },
```
giorgiosironi, line 12: first time I see a object key being an array? (if that's what it is)

thewilkybarkid: Which part? If you mean the property name, it's escaped as can't declare it with a `.` inside.

src/routes/entry-point.ts
thewilkybarkid: Needs a setting (see https://koajs.com/#settings).

.travis/smoke-test.sh
```diff
@@ -9,5 +9,9 @@ trap finish EXIT
 
 make start wait-healthy
 
-ping=$(curl -sS 'http://localhost:8080/' 2>&1)
-[[ $ping == 'OK' ]]
+output=$(curl --silent --show-error --write-out '%{http_code}' 'http://localhost:8080/')
+body=${output::${#output}-3}
```
giorgiosironi: very obscure; `echo "$output" | head -n 1` and `echo "$output" | tail -n 1`?

src/routes/entry-point.ts
giorgiosironi: Difference with `'schema.name': ...`?

giorgiosironi: Ah it's a lookup, would require another assignment then like `let schemaName = schema.name;`

.travis/smoke-test.sh
thewilkybarkid: Can't assume the body is one line.

giorgiosironi: Ah, meant `head -n -1` (print all but the last 1 line[s])

test/routes/entry-point.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import createContext from './context';
+
+describe('entry-point', (): void => {
+  it('should return the entry point', async (): Promise<void> => {
+    const context = createContext('entry-point');
+    const next = jest.fn();
+
+    await entryPoint(context.router)(context, next);
+
+    const { response } = context;
```
giorgiosironi: this part is verbose, compared to the scope of executing a single request. Extract an abstraction?

src/routes/entry-point.ts
```diff
@@ -0,0 +1,18 @@
...
+export default (router: Router): Middleware => (
+  async ({ request, response }: Context, next: Next): Promise<void> => {
+    response.body = {
+      '@id': url.resolve(request.origin, router.url(Routes.EntryPoint, {})),
+      '@type': schema.EntryPoint,
+      [schema.name]: { '@value': 'Article Store', '@language': 'en' },
+    };
+    response.type = 'jsonld';
+
+    await next();
```
giorgiosironi, line 16: annoying to have to remember to call `next()`; what is the use case `@koa/router` solves with this?

thewilkybarkid: Having middleware on specific routes, rather than globally. The example (https://github.com/koajs/router/blob/master/API.md#multiple-middleware) isn't particurlarly realistic. But since it's possible probably good practice to include it.

test/middleware.ts
```diff
@@ -0,0 +1,11 @@
+import {
+  Context, Middleware, Next, Response,
+} from 'koa';
```
thewilkybarkid, line 3: Not sure why eslint wants it like this?

test/routes/entry-point.test.ts
thewilkybarkid: Attempted in https://github.com/libero/article-store/pull/10/commits/a86ac8f0f1423d6f0315af53cbcb45a9c8cc37cf.

test/middleware.ts
thewilkybarkid: Narrowed it down to https://github.com/airbnb/javascript/blob/2e3adc98c784be7fa63c7a05271f7aa86c9d6bc7/packages/eslint-config-airbnb-base/rules/style.js#L399. Something to look at when we create a Libero config.

diversemix: I agree with https://github.com/airbnb/javascript/issues/1051#issuecomment-244526891

src/routes/entry-point.ts
diversemix: > annoying to have to remember to call `next()`; what is the use case `@koa/router` solves with this?

On the contrary - it's good practice as it gives the the option to fail fast on the request and stop passing it along. In contrast to `python` based frameworks where using expections is the pythonic way :smile:

giorgiosironi (approved):

PR #11, opened 2019-11-13
Support cross-origin requests

thewilkybarkid:
Allows any cross-origin request to be made against the API.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -7,9 +7,7 @@ const app = new Koa();
 const router = createRouter();
 
 app.use(logger());
-app.use(cors({
-  origin: '*',
-}));
+app.use(cors());
 app.use(router.routes());
 app.use(router.allowedMethods({ throw: true }));
 
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,3 +1,4 @@
+import cors from '@koa/cors';
 import Koa from 'koa';
 import logger from 'koa-logger';
 import createRouter from './router';
@@ -6,6 +7,9 @@ const app = new Koa();
 const router = createRouter();
 
 app.use(logger());
+app.use(cors({
+  origin: '*',
+}));
 app.use(router.routes());
 app.use(router.allowedMethods({ throw: true }));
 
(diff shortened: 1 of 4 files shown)


test/app.test.ts
```diff
@@ -13,4 +13,11 @@ describe('the application', (): void => {
 
     expect(response.status).toEqual(404);
   });
+
+  it('should support cross-origin requests', async (): Promise<void> => {
+    const response = await request(app.callback()).get('/').set('Origin', 'http://example.com');
+
+    expect(response.get('Access-Control-Allow-Origin')).toBe('*');
+    expect(response.get('Vary')).toContain('Origin');
```
thewilkybarkid: Need to check if this is an array, or still a comma-separated string.

src/app.ts
```diff
@@ -6,6 +7,9 @@ const app = new Koa();
 const router = createRouter();
 
 app.use(logger());
+app.use(cors({
+  origin: '*',
```
stephenwf: Was putting the origin as `*` meant to be a safe-guard against allowing cross-origin authorisation requests? If the store needs to deal with JWTs or any `Authorization` header values the `Access-Control-Allow-Origin` header _must_ match the `Origin` header from browsers. Setting the origin to this value is the default behaviour in koa cors (i.e. `app.use(cors())`). Just wanted to check if this value was on purpose (and if we should leave a comment?)

test/app.test.ts
giorgiosironi: It varies only on the presence or absence of `Origin` right?

src/app.ts
thewilkybarkid: Ah yeah, that's the point of `Vary: Origin`.

We will need authorization later down the line.

test/app.test.ts
thewilkybarkid: No, value of...

thewilkybarkid: Relates to https://github.com/libero/article-store/pull/11#discussion_r345751382.

giorgiosironi (approved):

PR #13, opened 2019-11-13
Use dummy router in middleware tests

thewilkybarkid:
Rather than declaring what routes this middleware depends on, this dummy implementation responds to any URL generatation requests.

### DIFF
--- a/test/routes/context.ts
+++ b/test/routes/context.ts
@@ -4,7 +4,7 @@ import Request from 'koa/lib/request';
 import Response from 'koa/lib/response';
 import { Request as IncomingMessage, Response as ServerResponse } from 'mock-http';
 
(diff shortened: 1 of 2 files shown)


test/routes/context.ts
```diff
@@ -4,16 +4,19 @@ import Request from 'koa/lib/request';
...
   response.req = request.req;
   response.res = new ServerResponse();
 
-  const router = new Router();
-  router.get(route, `path-to/${route}`, jest.fn());
+  const router = {
+    url(name: string): string {
+      return `/path-to/${name}`;
+    },
+  } as unknown as Router;
```
giorgiosironi, line 19: bit obscure but it's just a (double) cast in the end

giorgiosironi (approved):

PR #14, opened 2019-11-13
Add Hydra API Documentation

thewilkybarkid:
Turns the entry point into a Hydra entry point.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,13 +3,12 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.2",
-    "http-link-header": "^1.0.2",
+    "format-link-header": "^2.1.0",
     "koa": "^2.11.0",
     "koa-logger": "^3.2.1",
     "rdf-namespaces": "^1.6.0"
   },
   "devDependencies": {
-    "@types/http-link-header": "^1.0.1",
     "@types/jest": "^24.0.22",
     "@types/jsonld": "^1.5.0",
     "@types/koa": "^2.0.52",
@@ -17,6 +16,7 @@
     "@types/koa__cors": "^2.2.3",
     "@types/koa__router": "^8.0.1",
     "@types/node": "^12.12.6",
+    "@types/parse-link-header": "^1.0.0",
     "@types/supertest": "^2.0.8",
     "@typescript-eslint/eslint-plugin": "^2.6.1",
     "@typescript-eslint/parser": "^2.6.1",
@@ -28,6 +28,7 @@
     "jsonld": "^1.8.1",
     "mock-http": "^1.0.2",
     "nodemon": "^1.19.4",
+    "parse-link-header": "^1.0.1",
     "supertest": "^4.0.2",
     "ts-jest": "^24.1.0",
     "ts-node": "^8.4.1",
--- a/package.json
+++ b/package.json
@@ -3,11 +3,13 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.2",
+    "http-link-header": "^1.0.2",
     "koa": "^2.11.0",
     "koa-logger": "^3.2.1",
-    "rdf-namespaces": "^1.5.0"
+    "rdf-namespaces": "^1.6.0"
   },
   "devDependencies": {
+    "@types/http-link-header": "^1.0.1",
     "@types/jest": "^24.0.22",
     "@types/jsonld": "^1.5.0",
     "@types/koa": "^2.0.52",
--- a/src/middleware/api-documentation.ts
+++ b/src/middleware/api-documentation.ts
@@ -2,15 +2,16 @@ import Router from '@koa/router';
 import formatLinkHeader from 'format-link-header';
 import { Context, Middleware, Next } from 'koa';
 import { hydra } from 'rdf-namespaces';
+import url from 'url';
 import Routes from '../routes';
 
 export default (router: Router): Middleware => (
-  async ({ response }: Context, next: Next): Promise<void> => {
+  async ({ request, response }: Context, next: Next): Promise<void> => {
     await next();
 
     const link = {
       rel: hydra.apiDocumentation,
-      url: router.url(Routes.ApiDocumentation, {}),
+      url: url.resolve(request.origin, router.url(Routes.ApiDocumentation, {})),
     };
 
     response.append('Link', formatLinkHeader({ hydra: link }));
--- a/src/middleware/api-documentation.ts
+++ b/src/middleware/api-documentation.ts
@@ -1,5 +1,5 @@
 import Router from '@koa/router';
-import Link from 'http-link-header';
+import formatLinkHeader from 'format-link-header';
 import { Context, Middleware, Next } from 'koa';
 import { hydra } from 'rdf-namespaces';
 import Routes from '../routes';
@@ -8,11 +8,11 @@ export default (router: Router): Middleware => (
   async ({ response }: Context, next: Next): Promise<void> => {
     await next();
 
-    const link = new Link().set({
+    const link = {
       rel: hydra.apiDocumentation,
-      uri: router.url(Routes.ApiDocumentation, {}),
-    });
+      url: router.url(Routes.ApiDocumentation, {}),
+    };
 
-    response.append('Link', link.toString());
+    response.append('Link', formatLinkHeader({ hydra: link }));
   }
 );
--- a/src/middleware/api-documentation.ts
+++ b/src/middleware/api-documentation.ts
@@ -1,0 +1,18 @@
+import Router from '@koa/router';
+import Link from 'http-link-header';
+import { Context, Middleware, Next } from 'koa';
+import { hydra } from 'rdf-namespaces';
+import Routes from '../routes';
+
+export default (router: Router): Middleware => (
+  async ({ response }: Context, next: Next): Promise<void> => {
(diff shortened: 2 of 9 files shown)


src/routes/api-documentation.ts
```diff
@@ -0,0 +1,38 @@
...
+          '@id': schema.EntryPoint,
+          '@type': hydra.Class,
+          [hydra.title]: 'API entry point',
+          [hydra.supportedOperation]: {
+            '@type': hydra.Operation,
+            [hydra.title]: 'Get the entry point',
+            [hydra.method]: 'GET',
+            [hydra.expects]: {
+              '@id': owl.Nothing,
+            },
```
thewilkybarkid: Seen elsewhere, but seems redundant?

src/routes/api-documentation.ts
```diff
@@ -0,0 +1,38 @@
...
+      },
+      [hydra.supportedClass]: [
+        {
+          '@id': schema.EntryPoint,
+          '@type': hydra.Class,
+          [hydra.title]: 'API entry point',
+          [hydra.supportedOperation]: {
+            '@type': hydra.Operation,
+            [hydra.title]: 'Get the entry point',
+            [hydra.method]: 'GET',
```
thewilkybarkid: Not sure if this is really needed?

src/routes/api-documentation.ts
```diff
@@ -0,0 +1,38 @@
...
+            [hydra.expects]: {
+              '@id': owl.Nothing,
+            },
+            [hydra.returns]: {
+              '@id': schema.EntryPoint,
+            },
+          },
```
thewilkybarkid: I haven't defined any `hydra:supportedProperty`, not sure there's much need? We'll be willing to pass a lot of unknown properties around, so feels like it's only worth declaring required (eg `schema:name`) and impossible (eg `schema:potentialActions`) properties. But then a generic Hydra UI can't produce a useful form...

(More refers to read-write entities.)

src/middleware/api-documentation.ts
```diff
@@ -0,0 +1,18 @@
...
+import Link from 'http-link-header';
+import { Context, Middleware, Next } from 'koa';
+import { hydra } from 'rdf-namespaces';
+import Routes from '../routes';
+
+export default (router: Router): Middleware => (
+  async ({ response }: Context, next: Next): Promise<void> => {
+    await next();
+
+    const link = new Link().set({
```
thewilkybarkid: Not the best constructor...

package.json
```diff
@@ -3,11 +3,13 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.2",
+    "http-link-header": "^1.0.2",
     "koa": "^2.11.0",
     "koa-logger": "^3.2.1",
-    "rdf-namespaces": "^1.5.0"
+    "rdf-namespaces": "^1.6.0"
```
tpluscode, line 11: You might also be interested in [`@zazuko/rdf-vocabularies`](https://www.npmjs.com/package/@zazuko/rdf-vocabularies) as an alternative.

src/middleware/api-documentation.ts
```diff
@@ -0,0 +1,18 @@
...
+      rel: hydra.apiDocumentation,
+      uri: router.url(Routes.ApiDocumentation, {}),
+    });
+
+    response.append('Link', link.toString());
```
tpluscode: Try [set-link](https://www.npmjs.com/package/set-link)

```suggestion
    response.setLink(router.url(Routes.ApiDocumentation, {}), hydra.apiDocumentation)
```

package.json
thewilkybarkid: Will take a look separately. (PRs always welcome of course. 😃)

test/app.test.ts
```diff
@@ -23,4 +23,10 @@ describe('the application', (): void => {
     expect(response.get('Access-Control-Allow-Origin')).toBe('http://example.com');
     expect(parseHeader(response.get('Vary'))).toContain('Origin');
   });
+
+  it('should have an API documentation link', async (): Promise<void> => {
+    const response = await request(app.callback()).get('/');
+
+    expect(response.get('Link')).toContain('rel=http://www.w3.org/ns/hydra/core#apiDocumentation');
```
giorgiosironi: should this `Link`'s presence be tested in the CORS context as well?

test/routes/api-documentation.test.ts
```diff
@@ -0,0 +1,40 @@
...
+  it('should return the API documentation', async (): Promise<void> => {
+    const response = await makeRequest();
+    const graph = await jsonld.expand(response.body);
+
+    expect(graph).toHaveLength(1);
+
+    const object = graph[0];
+
+    expect(object['@id']).toBe('http://example.com/path-to/api-documentation');
+    expect(object['@type']).toContain('http://www.w3.org/ns/hydra/core#ApiDocumentation');
```
giorgiosironi, line 30: meaning the `@type` is an array after expansion?

thewilkybarkid: Yes, can be more than one.

src/middleware/api-documentation.ts
thewilkybarkid: Had a look, have to let TypeScript know about the new method so gone with yet another library in https://github.com/libero/article-store/pull/14/commits/d23e4347f483c3ffe76b0c34bc6ba6947b53033e.

test/app.test.ts
thewilkybarkid: Good catch. https://github.com/libero/article-store/pull/14/commits/148c7e23b3d9efe396385e63e311318be413908a

src/middleware/api-documentation.ts
tpluscode: How does express behave when you call `append('Link')` multiple times? `set-link` nicely merges multiple calls.

If typings are an issue, you may enjoy DefinitelyTyped/DefinitelyTyped#40374

thewilkybarkid: Not sure, this is Koa. 😅

(Not sure how Koa does it either. Probably the same.)

thewilkybarkid: Would prefer a syntax like `set-link` though. Maybe can refactor later.

giorgiosironi (commented):
From the Node.js point of view nothing to change, not an expert on the Hydra semantics

package.json
```diff
@@ -3,11 +3,12 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.2",
+    "format-link-header": "^2.1.0",
```
thewilkybarkid, line 6: Opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40411 for typing.

tpluscode: FYI `set-link` now also has typings ;)

src/routes/api-documentation.ts
```diff
@@ -0,0 +1,44 @@
...
+      [hydra.supportedClass]: [
+        {
+          '@id': schema.EntryPoint,
+          '@type': hydra.Class,
+          [hydra.title]: { '@value': 'API entry point', '@language': 'en' },
+          [hydra.supportedOperation]: {
+            '@type': hydra.Operation,
+            [hydra.title]: { '@value': 'Get the entry point', '@language': 'en' },
+            [hydra.method]: { '@value': 'GET' },
+            [hydra.expects]: { '@id': owl.Nothing },
```
alien-mcl, line 24: I find it ... unusual. While this is not a mistake, i feel that some clients could be confused as OWL may not be understood correctly. Hydra's [Heracles.ts](https://github.com/HydraCG/Heracles.ts) reference client will end up with a collection of `owl:Nothing` classes that are expected. I think it will do more harm than good - preferred solution would be to have no `hydra:expects` predicate here at all.

src/routes/api-documentation.ts
```diff
@@ -0,0 +1,44 @@
...
+      },
+      '@id': router.url(Routes.ApiDocumentation),
+      '@type': hydra.ApiDocumentation,
+      [hydra.entrypoint]: { '@id': router.url(Routes.EntryPoint) },
+      [hydra.supportedClass]: [
+        {
+          '@id': schema.EntryPoint,
+          '@type': hydra.Class,
+          [hydra.title]: { '@value': 'API entry point', '@language': 'en' },
+          [hydra.supportedOperation]: {
```
alien-mcl, line 20: Providing this supported operation is an overkill, I think. Providing a `hydra:entrypoint` predicate is enough as it is a `hydra:Link`.

alien-mcl (commented):
I'm really happy to see hydra powering libero. Feel free to ask any questions and features!

src/routes/api-documentation.ts
tpluscode: I know your OWL allergy but this may be the best explicit way to state that an operation really does not want a payload. This is also how the original issue tracker example was implemented I think

alien-mcl: I'm still not convinced - I feel i.e. empty RDF list still would be better, but I won't die for it. Just saying that OWL is not widely understood on client side

tpluscode: What is not to understand? I literally says "expects nothing" 🤣

alien-mcl: Indeed - client that does not know what _owl_ is it will try to create instances of class `owl:Nothing`!

PR #15, opened 2019-11-13
Follow official Node.js Docker image best practices

giorgiosironi:
From https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#non-root-user:
 
- [x] Run as `node` user prevents
  - privilege escalation to the host in case of Docker vulnerability
  - write access to all of the container's files from JavaScript code
- [x] Document that `npm run` is acting as PID 1, handling signals to stop the application

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -76,4 +76,6 @@ ENV NODE_ENV=production
 COPY --from=npm-prod /app/ .
 COPY --from=build-prod /app/build/ build/
 
+USER node
+
 CMD ["npm", "run", "start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,4 +78,5 @@ COPY --from=build-prod /app/build/ build/
 
 USER node
 
+# npm acts as PID 1, handling signals to stop the application immediately
 CMD ["npm", "run", "start"]


giorgiosironi:
Added https://github.com/orgs/libero/teams/infrastructure-reviewers with `Read` permissions for review requests.

Dockerfile
```diff
@@ -76,4 +76,7 @@ ENV NODE_ENV=production
 COPY --from=npm-prod /app/ .
 COPY --from=build-prod /app/build/ build/
 
+USER node
```
thewilkybarkid, line 79: Guess we do want this after things have been copied in?

thewilkybarkid: And no need to replicate on `dev`.

thewilkybarkid: Just read the guidance, makes sense to only include it at the end of `prod`.

thewilkybarkid (approved):

Dockerfile
giorgiosironi: `COPY` doesn't follow `USER` for file ownership, but `RUN npm ...` will; so my default is to switch at the end as I'm not very concerned about `root` in the build phase. Can be moved up to just after `npm run build` but not very useful to do so.

PR #16, opened 2019-11-13
Compact JSON-LD responses

thewilkybarkid:
Reduces the size of the responses by setting a default vocabulary and language in the context.


src/middleware/jsonld.ts
```diff
@@ -0,0 +1,20 @@
+import jsonld from 'jsonld';
+import { Context as JsonLdContext } from 'jsonld/jsonld-spec';
```
thewilkybarkid, line 3: This is a TypeScript-only file, eslint can't find it so must be configured wrong.

src/middleware/jsonld.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export default (context: JsonLdContext): Middleware => (
+  async ({ response }: Context, next: Next): Promise<void> => {
+    await next();
+
+    if (!isJsonLd(response)) {
+      return;
+    }
+
+    response.body = await jsonld.compact(response.body, context);
```
giorgiosironi, line 23: is compacting of universal utility or should it be skipped when using the development target (like minification and similar)?

thewilkybarkid: I think it's useful generally. It will be replaced with profile negotation at some point too.

src/middleware/jsonld.ts
thewilkybarkid: https://github.com/libero/article-store/pull/17

src/app.ts
```diff
@@ -13,6 +14,14 @@ app.use(logger());
...
   exposeHeaders: ['Link'],
 }));
+app.use(jsonld({
+  '@language': 'en',
+  '@vocab': 'http://schema.org/',
+  hydra: 'http://www.w3.org/ns/hydra/core#',
+  owl: 'http://www.w3.org/2002/07/owl#',
+  rdf: 'http://www.w3.org/1999/02/22-rdf-syntax-ns#',
+  rdfs: 'http://www.w3.org/2000/01/rdf-schema#',
+}));
```
thewilkybarkid: This isn't ideal, as introduces prefixes that might be unused, but the default might not also be appropriate (eg API documentation has no Schema.org).

test/middleware/jsonld.test.ts
```diff
@@ -0,0 +1,107 @@
...
+
+const makeRequest = async (next?: Next, jsonLdContext: JsonLdContext = {}): Promise<Response> => {
+  const context = createContext();
+
+  return runMiddleware(jsonld(jsonLdContext), context, next);
+};
+
+const next = (body: unknown, type: string) => async ({ response }: Context): Promise<void> => {
+  response.body = body;
+  response.type = type;
```
giorgiosironi, line 14: weird that it is the `next` middleware setting the body, although it checks that the `jsonld` middleware waits on it. If `response` was correctly populated at the start of the test this dummy middleware would be a simpler empty function?

test/middleware/jsonld.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    expect(contentType.parameters).toMatchObject({ profile: 'http://www.w3.org/ns/json-ld#compacted' });
+    expect(response.body).toMatchObject(expected);
+  });
+
+  it('should compact JSON-LD with a new context', async (): Promise<void> => {
+    const context = {
+      '@base': 'http://example.com',
+      '@language': 'en',
+      '@vocab': 'http://schema.org/',
+      dc: 'http://purl.org/dc/elements/1.1/',
```
giorgiosironi, line 53: inconsistent that these keys are unquoted, if they are meant to be used literally as `dc` or `name`?

test/middleware/jsonld.test.ts
thewilkybarkid: Still need one case to make sure it runs after `next`.

test/middleware/jsonld.test.ts
```diff
@@ -0,0 +1,107 @@
...
+    const contentType = parseContentType(response);
+
+    expect(response.type).toBe('application/ld+json');
+    expect(contentType.type).toBe('application/ld+json');
+    expect(contentType.parameters).toMatchObject({});
+    expect(response.body).toBe('text');
+  });
+
+  it('should do nothing if the response JSON-LD but there is no body', async (): Promise<void> => {
+    const response = await makeRequest(next(undefined, 'jsonld'));
```
giorgiosironi, line 85: if there is no body how is the response JSON-LD or even JSON? What is the use case?

test/middleware/jsonld.test.ts
thewilkybarkid: Yep, JS object and coding standard. (Doesn't help that a lot of the keys contain special characters.)

test/middleware/jsonld.test.ts
thewilkybarkid: Edge case, and definitely shouldn't happen inside the app, but `Content-Type` headers can lie...

test/middleware/jsonld.test.ts
giorgiosironi: Yes, but if that part breaks you get 1 failing test instead of all of them.

thewilkybarkid: Understand; think changing it doesn't make it simpler though. Happy if you'd like to refactor.

giorgiosironi: Not much value in changing towards either way for me, raised just for the sake of discussion.

giorgiosironi (approved):
Has conflicts to solve, but no further concerns

PR #18, opened 2019-11-14
Throw a Not Found error

thewilkybarkid:
Currently an unmatched route ends up with the status code being `404` rather than an error thrown. This converts any error status code (where no body has been set) to a `HttpError` and throws it. This will allow all errors to be caught consistently, but has no visible effect right now.


thewilkybarkid:
(Feature or refactor, given there's no visible change? 🤔)

diversemix:
Could you give a few more words why this is being done?

thewilkybarkid:
As part of https://github.com/libero/publisher/issues/314 we'll be making sure error responses are in the Hydra format. Currently 404s aren't thrown errors, so need to handle both thrown errors and recognise a response with a 404 status code and no body.

thewilkybarkid:
Other option could be to make `@koa/router` throw a 404 if no routes match...

diversemix (approved):

src/middleware/http-errors.ts
```diff
@@ -0,0 +1,13 @@
+import { Context, Middleware, Next } from 'koa';
```
giorgiosironi: `middleware/http-errors` has the same base name as the library `http-errors`, possible confusion when looking at the imported module in the middle of other files. Is it the most fitting name?

thewilkybarkid:
I'm going to simplify this, in favour of getting the router to throw a `NotFound` error on a 404.

src/middleware/routing.ts
```diff
@@ -0,0 +1,22 @@
...
+    if (typeof _matchedRoute === 'undefined') {
+      throw new createHttpError.NotFound();
+    }
+  }
+);
+
+export default (router: Router): Middleware => (
+  compose([
+    router.routes(),
+    notFound(),
```
thewilkybarkid, line 19: Will propose changing the library to allow `router.routes({ throw: true })`.

giorgiosironi (approved):

PR #19, opened 2019-11-14
Return errors in Hydra format

thewilkybarkid:
Converts all errors into the Hydra error format.

Depends on #18.


package.json
```diff
@@ -7,7 +7,7 @@
     "koa": "^2.11.0",
     "koa-compose": "^4.1.0",
     "koa-logger": "^3.2.1",
-    "rdf-namespaces": "^1.5.0"
+    "rdf-namespaces": "^1.6.0"
```
giorgiosironi: ah, https://gitlab.com/vincenttunru/rdf-namespaces/merge_requests/3

src/middleware/error-handler.ts
```diff
@@ -0,0 +1,33 @@
+import createError, { HttpError } from 'http-errors';
+import { Context, Middleware, Next } from 'koa';
+import { hydra } from 'rdf-namespaces';
+
+const handleHttpError = (error: HttpError, { response }: Context): void => {
+  response.status = error.status;
+  response.body = {
+    '@type': hydra.Status,
+    [hydra.title]: { '@value': response.message, '@language': 'en' },
```
giorgiosironi, line 9: at first `response.message` looked magic in comparison to `error.message`, but thanks to the test now clear it's the `Not Found`|`Internal Server Error`|... string.

src/middleware/error-handler.ts
```diff
@@ -0,0 +1,33 @@
...
+  };
+
+  if (error.message !== response.message) {
```
giorgiosironi, line 12: seems a weak check to be performed on error strings rather than on what the tests describe, which is the exception being a `non-http error`. But the type of error is `HttpError` so not sure how it got here in the first place.

test/middleware/error-handler.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    });
+
+    it('should return details about the error', async (): Promise<void> => {
+      const response = await makeRequest(next('Some Error'));
+      const object = await jsonld.compact(response.body, { '@language': 'en' });
+
+      expect(object).not.toHaveProperty('@id');
+      expect(object['@type']).toBe('http://www.w3.org/ns/hydra/core#Status');
+      expect(object['http://www.w3.org/ns/hydra/core#title']).toBe('Internal Server Error');
+      expect(object['http://www.w3.org/ns/hydra/core#description']).toBe('Some Error');
```
giorgiosironi, line 73: verbosity is good here as the field names are unambiguous

src/middleware/error-handler.ts
giorgiosironi: Ah there's a `toHttpError`. Maybe leave the original thrown object rather than wrapping in `HttpError`, or mark it so that it is clear it wasn't originally if it has an option for that? Otherwise I suspect throwing a 500 is indistinguishable from throwing a string (from this middleware rendering a response).

thewilkybarkid: Would you ever throw a 500?

giorgiosironi: If there isn't a clear mapping of the server-side error to a status code; but yes, why bothering when you can just throw the right exception instead.

giorgiosironi (approved):

PR #20, opened 2019-11-15
Remove empty URL parameters

thewilkybarkid:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40360 was merged, so the definition of `Router.url()` now correctly has the `params` argument as optional.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "@types/koa-compose": "^3.2.5",
     "@types/koa-logger": "^3.1.1",
     "@types/koa__cors": "^2.2.3",
-    "@types/koa__router": "^8.0.1",
+    "@types/koa__router": "^8.0.2",
     "@types/node": "^12.12.6",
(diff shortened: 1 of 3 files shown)


giorgiosironi (approved):

package.json
```diff
@@ -17,7 +17,7 @@
     "@types/koa-compose": "^3.2.5",
     "@types/koa-logger": "^3.1.1",
     "@types/koa__cors": "^2.2.3",
-    "@types/koa__router": "^8.0.1",
+    "@types/koa__router": "^8.0.2",
```
giorgiosironi, line 20: weird naming convention, I also found the oldest `@types/koa-router` which is probably abandoned

thewilkybarkid: Double underscore is for the namespaced convention:

- `@types/koa-router` is for `koa-router`
- `@types/koa__router` is for `@koa/router`

PR #21, opened 2019-11-15
Add Hydra console

thewilkybarkid:
Runs a Generic Hypermedia App container to access the API through.

### DIFF
--- a/.docker/docker-compose.yml
+++ b/.docker/docker-compose.yml
@@ -9,8 +9,8 @@ services:
   console:
     image: tpluscode/hypermedia-app:version-0.12.1
     environment:
-      - 'API_ENTRYPOINTS={ "http://localhost:8080/": "Article Store" }'
-      - 'BASE_URL=http://localhost:8080/'
+      API_ENTRYPOINTS: '{ "http://localhost:8080/": "Article Store" }'
+      BASE_URL: http://localhost:8080/
     ports:
       - '8081:8000'
     depends_on:
--- a/.docker/docker-compose.yml
+++ b/.docker/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     ports:
       - '8080:8080'
   console:
-    image: tpluscode/hypermedia-app:version-0.12.0
+    image: tpluscode/hypermedia-app:version-0.12.1
     environment:
       - 'API_ENTRYPOINTS={ "http://localhost:8080/": "Article Store" }'
       - 'BASE_URL=http://localhost:8080/'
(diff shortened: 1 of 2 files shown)


.docker/docker-compose.yml
```diff
@@ -6,3 +6,12 @@ services:
       context: ../
     ports:
       - '8080:8080'
+  console:
+    image: tpluscode/hypermedia-app:alpha
+    environment:
+      - 'API_ENTRYPOINTS={ "http://localhost:8080/": "Article Store" }'
+      - 'BASE_URL=http://localhost:8080/'
```
thewilkybarkid: Not working due to https://github.com/hypermedia-app/generic.hypermedia.app/issues/44.

.docker/docker-compose.yml
```diff
@@ -6,3 +6,12 @@ services:
       context: ../
     ports:
       - '8080:8080'
+  console:
+    image: tpluscode/hypermedia-app:alpha
```
thewilkybarkid: Will need a tag released before can merge. /cc @tpluscode

thewilkybarkid: Heh, hadn't noticed https://hub.docker.com/layers/tpluscode/hypermedia-app/version-0.12.0/images/sha256-dc1a9ddf92da97461275fd819a0e5029ae8203027f4e3f1e9205ff5c8b9227b9 was pushed. 🎉

tpluscode: Unfortunately it's still one with the environment variables not honoured (hypermedia-app/generic.hypermedia.app#44). Stay tuned for build 0.12.1

tpluscode:
~~By the way, please note that the `BASE_URL` is a base of the called API and not the client application.~~

~~The container exposes port `8000` (unchanged default of `local-web-server`).~~

never mind

.docker/docker-compose.yml
tpluscode: > docker pull tpluscode/hypermedia-app:version-0.12.1

stephenwf (approved):
Not so keen on unescaped JSON as environment variables, I assume docker does something clever to escape them?
```
- 'API_ENTRYPOINTS={ "http://localhost:8080/": "Article Store" }'
```
Can see from the docs that how it should be configured though.

tpluscode:
> Not so keen on unescaped JSON as environment variables, I assume docker does something clever to escape them?
> 
> ```
> - 'API_ENTRYPOINTS={ "http://localhost:8080/": "Article Store" }'
> ```
> 
> Can see from the docs that how it should be configured though.

While this is a set of key/value pairs, I found JSON to be the simplest to consume. I considered a double-delimited string like `api1=link1;api2=link2`. Let me know if you think that would be better?

PR #22, opened 2019-11-15
Add empty list of articles

thewilkybarkid:
Adds an empty collection of articles accessible from the entry point.

### DIFF
--- a/src/routes/api-documentation.ts
+++ b/src/routes/api-documentation.ts
@@ -1,6 +1,8 @@
 import Router from '@koa/router';
 import { Context, Middleware, Next } from 'koa';
-import { hydra, owl, schema } from 'rdf-namespaces';
+import {
+  hydra, owl, rdfs, schema,
+} from 'rdf-namespaces';
 import Routes from './index';
 
 export default (router: Router): Middleware => (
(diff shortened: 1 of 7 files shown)


src/routes/article-list.ts
```diff
@@ -0,0 +1,26 @@
...
+    response.body = {
+      '@context': {
+        '@base': request.origin,
+      },
+      '@id': router.url(Routes.ArticleList),
+      '@type': hydra.Collection,
+      [hydra.title]: { '@value': 'List of articles', '@language': 'en' },
+      'http://www.w3.org/ns/hydra/core#manages': {
+        'http://www.w3.org/ns/hydra/core#property': rdf.type,
+        'http://www.w3.org/ns/hydra/core#object': schema.Article,
```
thewilkybarkid, line 17: `hydra.manages` etc not available due to https://gitlab.com/vincenttunru/rdf-namespaces/merge_requests/3#note_243892161.

src/routes/entry-point.ts
```diff
@@ -12,6 +12,15 @@ export default (router: Router): Middleware => (
...
       [schema.name]: { '@value': 'Article Store', '@language': 'en' },
+      [hydra.collection]: {
+        '@id': router.url(Routes.ArticleList),
+        '@type': hydra.Collection,
+        [hydra.title]: { '@value': 'List of articles', '@language': 'en' },
+        'http://www.w3.org/ns/hydra/core#manages': {
+          'http://www.w3.org/ns/hydra/core#property': rdf.type,
+          'http://www.w3.org/ns/hydra/core#object': schema.Article,
+        },
+      },
```
thewilkybarkid: Want to avoid duplicating this.

src/routes/article-list.ts
stephenwf: We could hold these in a separate file and use something like `hydraExtra.manages` etc to avoid the duplication and make it easier to swap out later on?

src/routes/api-documentation.ts
```diff
@@ -35,6 +37,52 @@ export default (router: Router): Middleware => (
             },
           ],
         },
+        {
```
giorgiosironi, line 40: so it's the https://www.w3.org/ns/hydra/core#apiDocumentation that would be used by a generic client to build a form? Rather than embedded controls in an API response like with https://schema.org/Action?

thewilkybarkid: It'll be a mix; a Hydra client merges them. API details that apply everywhere go into the documentation, resource-specific details go into the resource itself (eg can update any article vs can update specific articles).

Related: https://httpapis.slack.com/archives/C1JP575EX/p1572601301095800

src/routes/article-list.ts
giorgiosironi: (after reading section 5.2 of the Hydra Core Vocabulary, it all checks out)

src/routes/article-list.ts
```diff
@@ -0,0 +1,26 @@
...
+      },
+      [hydra.totalItems]: 0,
+      [hydra.member]: { '@list': [] },
```
giorgiosironi, line 20: is the `@list` for ordering?

thewilkybarkid: Yep: https://w3c.github.io/json-ld-syntax/#lists

PR #23, opened 2019-11-16
Add type definitions for format-link-header

thewilkybarkid:
Follow up to https://github.com/libero/article-store/pull/14#discussion_r346762504.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -562,6 +562,12 @@
         "@types/range-parser": "*"
       }
     },
+    "@types/format-link-header": {
+      "version": "2.1.0",
+      "resolved": "https://registry.npmjs.org/@types/format-link-header/-/format-link-header-2.1.0.tgz",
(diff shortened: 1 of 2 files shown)


stephenwf:
How did this avoid breaking the build before this addition? Is there an additional `ts-ignore` or `any` that needs removed too?

thewilkybarkid:
Only usage is https://github.com/libero/article-store/blob/221d0391a66207213062e9c5daf398f8cb1d4110/src/middleware/api-documentation.ts#L12-L17

Wonder if there’s a way to require that dependencies have a type definition? (Since `any` can’t be stopped there.)

stephenwf:
Adding `"noImplicitAny": true` to the compiler options will do that. Running that on master yields:
```
▶ npx tsc
src/middleware/api-documentation.ts:2:30 - error TS7016: Could not find a declaration file for module 'format-link-header'. '/Users/stephen/github.com/libero/article-store/node_modules/format-link-header/src/index.js' implicitly has an 'any' type.
  Try `npm install @types/format-link-header` if it exists or add a new declaration (.d.ts) file containing `declare module 'format-link-header';`

2 import formatLinkHeader from 'format-link-header';
                               ~~~~~~~~~~~~~~~~~~~~


Found 1 error.
```
which looks about right to catch this one

edit: Added change to #25 if you want to add it 👍

stephenwf (approved):

PR #24, opened 2019-11-16
Add API documentation middleware test

thewilkybarkid:


test/middleware/api-documentation.test.ts
```diff
@@ -0,0 +1,25 @@
...
+  it('adds the API documentation link', async (): Promise<void> => {
+    const response = await makeRequest();
+
+    const links = parseLinkHeader(response.get('Link'));
+
+    expect(Object.keys(links)).toHaveLength(1);
+
+    const link = links['http://www.w3.org/ns/hydra/core#apiDocumentation'];
+
+    expect(link.url).toBe('http://example.com/path-to/api-documentation');
```
stephenwf: If this test fails and `link` is `undefined` it will throw an `Uncaught TypeError: Cannot read property of undefined` instead of a failing assertion. It might be easier to compact this into an inline snapshot (`expect(links).toMatchInlineSnapshot()`) or, if its not a getter/setter based object, just asserting the shape of the whole links object.

stephenwf (approved):

PR #26, opened 2019-11-18
chore: Fixed NPM Advisory for Handlebars dependency

stephenwf:
https://npmjs.com/advisories/1316

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3403,9 +3403,9 @@
       "dev": true
     },
     "handlebars": {
-      "version": "4.5.1",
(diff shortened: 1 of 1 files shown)


thewilkybarkid:
Setting up Dependabot to try and take care of these, no sign of a PR yet though.

thewilkybarkid:
Unclear what database Dependabot/GitHub are using for theirs advisories, bit concerning that this hasn't been flagged by either (though it's a dev-only dependency in something that we don't use, so 🤷‍♂).

thewilkybarkid (approved):

PR #27, opened 2019-11-18
Populated generics with RouterContext

stephenwf:


src/middleware/routing.ts
```diff
@@ -1,10 +1,10 @@
 import Router, { RouterContext } from '@koa/router';
 import createHttpError from 'http-errors';
-import { Middleware, Next } from 'koa';
+import { Middleware, Next, DefaultContext } from 'koa';
 import compose from 'koa-compose';
 
-const notFound = (): Middleware => (
-  async ({ _matchedRoute }: RouterContext, next: Next): Promise<void> => {
+const notFound = (): Middleware<DefaultContext, RouterContext> => (
```
thewilkybarkid: `DefaultState` not `DefaultContext`?

src/middleware/routing.ts
```diff
@@ -1,10 +1,10 @@
 import Router, { RouterContext } from '@koa/router';
 import createHttpError from 'http-errors';
-import { Middleware, Next } from 'koa';
+import { Middleware, Next, DefaultContext } from 'koa';
 import compose from 'koa-compose';
 
-const notFound = (): Middleware => (
-  async ({ _matchedRoute }: RouterContext, next: Next): Promise<void> => {
```
thewilkybarkid, line 7: Hmm, `Next` can also be inferred, but had been adding for clarity. Need to find the right balance...

giorgiosironi (commented):
:coffee:

src/middleware/routing.ts
stephenwf: Right, good spot!

src/middleware/routing.ts
stephenwf: I can remove it if you like? Removed the `{ _matchedRoute }` one since the type inferred is more complete:
Screenshot 2019-11-18 at 10 57 49

thewilkybarkid: Let's leave both in for now, with consistency with the others. Can have a think separately about it.

thewilkybarkid (approved):

thewilkybarkid:
Just spotted this will create a typing problem in `routing.test.ts` when `strict` is on.

stephenwf:
If we add the following to jest.config.js
```
globals: {
    'ts-jest': {
      tsConfig: {
        strict: false,
      },
    },
  },
```

that will disable strict mode just for tests (which should streamline creating tests).

thewilkybarkid:
Hmm, not ideal. Trying to work out how to fix but going round in circles.

We'll have a similar problem with use of the `jsonld` library too.

thewilkybarkid:
I _think_ https://github.com/libero/article-store/blob/aafad4b7e1c7a90046e2e9bbf916c0f21b765dad/test/middleware.ts#L5 needs to be:

```typescript
async (middleware: Middleware<DefaultState, RouterContext>, context: ParameterizedContext<DefaultState, RouterContext>, next?: Next): Promise<Response>
```

or even:

```typescript
async <StateT = DefaultState, CustomT = RouterContext>(middleware: Middleware<StateT, CustomT>, context: ParameterizedContext<StateT, CustomT>, next?: Next): Promise<Response>
```

thewilkybarkid (approved):
(If you're also happy!)

PR #28, opened 2019-11-18
Added type for unknown error, based on input of http-errors library

stephenwf:
This gives the `unknown` type a concrete type before being passed to the create error.

The type for UnknownError is computed as:
```typescript
type UnknownError = Error | string | number | { [key: string]: any };
```
From `createHttpError.CreateHttpError` types package. As this is not exported, there is a quick `infer` type that will grab this value, even if it changes.

```typescript
// `T extends ( ...` indicates that there is a ternary
type UnknownError<T = CreateHttpError> = T extends (
    // This expression is a generic shape of an input (in this case `CreateHttpError`)
    //  The `infer` keyword determines the true-ness of the statement
    (event: infer ErrorShape, ...args: unknown[]) => HttpError
  // When true, the value of the type is `ErrorShape` (a generic value, inferred above)
  // When false, the value is never and will error (same as the return type when throwing an exception)
  ) ? ErrorShape : never;
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -575,9 +575,9 @@
       "dev": true
     },
     "@types/http-errors": {
-      "version": "1.6.2",
-      "resolved": "https://registry.npmjs.org/@types/http-errors/-/http-errors-1.6.2.tgz",
-      "integrity": "sha512-mzJX3tIbtadNZQIDbfA9eW+mAjww7GY7WYcfKDGB5SSXMAzI8KD+5fvyX5FqcKtns346+WGwAJJ8cPsDxMz0lw==",
+      "version": "1.6.3",
+      "resolved": "https://registry.npmjs.org/@types/http-errors/-/http-errors-1.6.3.tgz",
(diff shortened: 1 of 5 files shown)


thewilkybarkid:
> As this is not exported

Sounds like it could be added upstream?

thewilkybarkid:
Opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40487.

stephenwf:
If we add the following to jest.config.js
```
globals: {
    'ts-jest': {
      tsConfig: {
        strict: false,
      },
    },
  },
```

that will disable strict mode just for tests (which should streamline creating tests).

thewilkybarkid:
Type change has been merged and released.

thewilkybarkid (approved):
(Assuming you're happy with the changes!)

stephenwf:
~Changed looks good, didn't spot the `UnknownError` type~ Re-read! :shipit:

PR #25, opened 2019-11-17
chore: Added no implicit any to compiler options

stephenwf:
Will prevent declarations (const, import, let, functional arguments etc.) being assigned to any when a type cannot be inferred. 

```typescript
function takesNumber(num) {
  // would throw compiler error
}
```

but will still allow inferred types, usually through generics:
```typescript
[1, 2, 3].map(num => {
  // num still inferred correctly as number
});

[1, 2, 'something'].map(num => {
  // num inferred as `number | string`
});
```

Depends on #23

PR #29, opened 2019-11-18
chore: Added "strict" value to compiler options in tsconfig

stephenwf:
Alternative to #25 with all strict options. Build fixed by #23, #27 and #28.


tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "es2019",
     "module": "commonjs",
     "outDir": "build",
-    "alwaysStrict": true,
```
thewilkybarkid, line 6: Is this not needed anymore?

stephenwf: Technically not, but could be useful to add the more verbose options for clarity. From the docs:
> Enable all strict type checking options.Enabling --strict enables --noImplicitAny, --noImplicitThis, --alwaysStrict, --strictBindCallApply, --strictNullChecks, --strictFunctionTypes and --strictPropertyInitialization.

thewilkybarkid:
Going to revisit after ~#64~ #115, as the jsonld library is incompatible.

thewilkybarkid:
Just tried apply to #115 and good news in that it's basically there.

thewilkybarkid:
After https://github.com/libero/article-store/pull/164, the remaining problems are purely related to dependency typings (1 without types, and a few case of using modules that aren't typed).

thewilkybarkid (approved):
That was quite a lot of work to satisfy a 1-line chore. 😅

PR #30, opened 2019-11-20
Configure Dependabot

thewilkybarkid:
Adds Dependabot configuration for updating the Dockerfile, NPM dependencies and Git submodules.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -12,7 +12,7 @@ update_configs:
 
   - package_manager: javascript
     directory: /
-    update_schedule: live
+    update_schedule: daily
     commit_message:
       prefix: chore
       include_scope: true
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,29 @@
+version: 1
+update_configs:
+
+  - package_manager: docker
+    directory: /
+    update_schedule: daily
+    commit_message:
+      prefix: chore
+      include_scope: true
+    default_labels:
+      - ':recycle: Chore'
+
+  - package_manager: javascript
+    directory: /
+    update_schedule: live
+    commit_message:
+      prefix: chore
+      include_scope: true
+    default_labels:
+      - ':recycle: Chore'
+
+  - package_manager: submodules
+    directory: /
+    update_schedule: daily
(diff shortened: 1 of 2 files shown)


thewilkybarkid:
No support yet for Docker Compose: https://github.com/dependabot/dependabot-core/pull/1431.

.dependabot/config.yml
```diff
@@ -0,0 +1,29 @@
...
+    update_schedule: daily
+    commit_message:
+      prefix: chore
+      include_scope: true
+    default_labels:
+      - ':recycle: Chore'
+
+  - package_manager: javascript
+    directory: /
+    update_schedule: live
```
thewilkybarkid: Might be too much.

.dependabot/config.yml
```diff
@@ -0,0 +1,29 @@
...
+update_configs:
+
+  - package_manager: docker
+    directory: /
+    update_schedule: daily
+    commit_message:
+      prefix: chore
+      include_scope: true
+    default_labels:
+      - ':recycle: Chore'
```
thewilkybarkid, line 11: Assume this is how you reference a label with an emoji. 🤞

.dependabot/config.yml
giorgiosironi: Like, everything ~every day~ all the time?

thewilkybarkid: Yeah, rather than getting a batch once a day, bigger batch once a week or large batch once a month. (Not sure how Dependabot handles update depths, if it only updates top-level dependencies then not much of a problem, but if there's individual PRs for _everything_ in `package-lock.json`...)

giorgiosironi: A common side-effect is that build time is spent on testing the updates, so needs batching at the moment to avoid Travis CI being a bottleneck; especially if there are many PRs open as they are all rebased continuously to resolve conflicts. If moving to Github Actions the bottleneck may become code review time.

giorgiosironi (approved):
Would like to see it in action and tune it

giorgiosironi:
Created https://github.com/libero/publisher/issues/340 to reduce the risk of Travis CI being a bottleneck later on.

PR #33, opened 2019-11-20
Add Hydra validator to CI

thewilkybarkid:
Add [hydra-validator-analyse](https://www.npmjs.com/package/hydra-validator-analyse) to the Article Store's CI, which should help find issues with the Hydra implementation.


.travis/analyse.sh
```diff
@@ -0,0 +1,12 @@
...
+
+function finish() {
+  make logs stop
+}
+
+trap finish EXIT
+
+make start wait-healthy
+
+make exec command="npx hydra-validator analyse http://localhost:8080/"
```
thewilkybarkid, line 12: Wonder if this is better Dockerised?

giorgiosironi: Trying to run it locally to check...

thewilkybarkid: Related: https://github.com/libero/article-store/pull/33/commits/53a41cc04fa63823c97e752605094f8f769ab544

giorgiosironi: Ah, it is Dockerized, using the application (`dev`) image

thewilkybarkid: Sorry, meant Dockerise the tool itself (rather than having it as an NPM dependency).

tpluscode: How would you propose to build the image and then use it?

giorgiosironi: `docker-compose` doesn't have jobs (in lieu of long-running services) so it's always a bit unnatural to include these as separate containers. We make do sometimes with a `ci` service with a `bash` command that exits with 0 during `docker-compose up`.

.travis.yml
```diff
@@ -13,5 +13,6 @@ install:
 script:
   - TARGET=prod .travis/smoke-test.sh
   - TARGET=dev .travis/smoke-test.sh
+  - .travis/analyse.sh
```
giorgiosironi: are these in an explicit order? If so static checks like `lint` should go before dynamic ones like tests and this?

.travis/analyse.sh
giorgiosironi:
```
docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.dev.yml exec app npx hydra-validator analyse htt
p://localhost:8080/
  INFO      Analysis started...  +0ms
  SUCCESS   Successfully fetched http://localhost:8080/  +0ms
  SUCCESS   -- Response status 200  +1ms
  SUCCESS   -- Api Documentation link found  +1ms
  SUCCESS   -- Successfully fetched http://localhost:8080/doc  +6ms
  SUCCESS   ---- Response status 200  +0ms
  SUCCESS   ---- Successfully parsed 55 triples  +20ms
  SUCCESS   ---- Entrypoint found: http://localhost:8080/  +1ms
  INFO      ------ Skipping already visited resource <http://localhost:8080/>  +47ms
  SUCCESS   ---- Found 2 Supported Classes  +1ms
  SUCCESS   -- Successfully parsed 3 triples  +1ms
  SUCCESS   Analysis complete Success: 9 +0ms
```

.travis.yml
thewilkybarkid: Not really. I'll move this to the end (reordering can happen elsewhere). But they do all run.

.travis/analyse.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -e
+
+function finish() {
+  make logs stop
+}
+
+trap finish EXIT
+
+make start wait-healthy
```
giorgiosironi, line 10: only thing to note was that this failed rather noisily when I was missing the submodule:
```
[14:06:04][giorgio@Newton:~/code/article-store]$ .travis/analyse.sh                                                    [98/251]
docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.dev.yml up --detach
Creating network "docker_default" with the default driver
Creating docker_app_1 ... done
Creating docker_console_1 ... done
services=($(docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.dev.yml ps --quiet)); \
if [ ${#services[@]} -eq 0 ]; then \
        echo "No containers running"; \
        exit 1; \
fi; \
for service in ${services}; do \
        .scripts/docker/wait-healthy.sh "${service}"; \
done
bash: line 6: .scripts/docker/wait-healthy.sh: No such file or directory
Makefile:34: recipe for target 'wait-healthy' failed
make: *** [wait-healthy] Error 127
docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.dev.yml logs
Attaching to docker_console_1, docker_app_1
app_1      | 
app_1      | > @ start:dev /app
app_1      | > nodemon --exec 'node --require ts-node/register ./src/index.ts' --ext .js,.json,.ts
app_1      | 
app_1      | [nodemon] 1.19.4
app_1      | [nodemon] to restart at any time, enter `rs`
app_1      | [nodemon] watching dir(s): *.*
app_1      | [nodemon] watching extensions: js,json,ts
app_1      | [nodemon] starting `node --require ts-node/register ./src/index.ts`
console_1  | Listening on http://68c32a97136c:8000, http://127.0.0.1:8000, http://192.168.144.3:8000
docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.dev.yml down
Stopping docker_console_1 ... 
Stopping docker_console_1 ... done
Stopping docker_app_1     ... done
Removing docker_console_1 ... done
Removing docker_app_1     ... done
Removing network docker_default
```
I wouldn't go as far as adding git commands to the Makefile, because CI systems like Travis or Jenkins take care of this automatically. But noting down a ` git submodule update -i` in the README could help. Once the submodule is initialized `git diff` will show updates so it will be impossible to miss.

thewilkybarkid: Probably should be in `make install` though.

giorgiosironi (approved):

.travis/analyse.sh
thewilkybarkid: Wouldn't want to build it locally/add it to Docker Compose, just have something like `docker run  --net=host tpluscode/hydra-validator analyse http://localhost:8080` here (or, better, in the Makefile).

giorgiosironi: If there is an image available (or can be built elsewhere) this can make this build faster and more stable, like you would download a Selenium container rather than installing [`selenium-standalone-jar`](https://www.npmjs.com/package/selenium-standalone-jar). The base questions are whether this application needs to load the project code and/or to run in the same process, the answer seems to be no to both.

giorgiosironi:
General opinion is only switch to a separate Docker container if there is an image already available. Benefit being a simpler and faster `npm install` process.

Otherwise create a separate issue tagged with [`infrastructure`](https://github.com/libero/publisher/issues?q=is%3Aissue+is%3Aopen+label%3Ainfrastructure) where we could contribute one upstream and then use it; we can prioritize that one at the next Replenishment.

PR #43, opened 2019-11-21
Allow creation of an article

nlisgo:
Expose an endpoint to allow the creation of an article with the only field being name.

Create a repository for the articles in memory and list the articles on the article list end point.


test/errors/article-has-no-id.test.ts
```diff
@@ -1,14 +1,14 @@
-import ArticleHasNoId from '../../src/errors/article-has-no-id';
+import ArticleHasNoID from '../../src/errors/article-has-no-id';
```
thewilkybarkid: Don't disagree, but reducing the diff in this PR.

src/routes/post-article.ts
```diff
@@ -0,0 +1,22 @@
...
+import { JsonLdObj } from 'jsonld/jsonld-spec';
+import uniqueString from 'unique-string';
+import Articles from '../articles';
+import ArticleIdAlreadySet from '../errors/article-id-already-set';
+
+export default (articles: Articles): Middleware => (
+  async ({ request, response }: Context, next: Next): Promise<void> => {
+    const article = request.body as JsonLdObj;
+    if ('@id' in article) {
+      throw new ArticleIdAlreadySet(article['@id']);
```
thewilkybarkid: Been thinking about this. `409 Conflict` makes sense if an article with that ID is already in the collection, but really we're saying you're not allowed to set an ID, so `403 Forbidden` covers it. Suggest we also throw an HTTP error directly here for simplicity.

nlisgo: I've made this refactor.

src/routes/post-article.ts
```diff
@@ -0,0 +1,22 @@
...
+import { JsonLdObj } from 'jsonld/jsonld-spec';
+import uniqueString from 'unique-string';
+import createHttpError from 'http-errors';
+import Articles from '../articles';
+
+export default (articles: Articles): Middleware => (
+  async ({ request, response }: Context, next: Next): Promise<void> => {
+    const article = request.body as JsonLdObj;
+    if ('@id' in article) {
+      throw new createHttpError.Forbidden('Article ID is not permitted');
```
nlisgo: Is there value is passing the id into the message for logging?

Perhaps: `Article ID is not permitted ('_:1' given)`

src/routes/index.ts
```diff
@@ -1,6 +1,7 @@
 enum Routes {
   'ApiDocumentation' = 'api-documentation',
   'ArticleList' = 'article-list',
+  'PostArticle' = 'post-article',
```
thewilkybarkid: Suggest we call it `CreateArticle`. Maybe even `AddArticle`?

nlisgo: Will go for `AddArticle` as it matches the method call.

src/routes/post-article.ts
thewilkybarkid: Yeah, maybe even `Article IDs must not be set ('_:1' was given)`.

thewilkybarkid:
I've added 3 tasks to https://github.com/libero/publisher/issues/291 (re validation), ok to not include in this PR (matches the API documentation) but we'll need a follow-up.

src/routes/post-article.ts
```diff
@@ -0,0 +1,22 @@
+import { constants } from 'http2';
+import { Context, Middleware, Next } from 'koa';
+import { JsonLdObj } from 'jsonld/jsonld-spec';
+import uniqueString from 'unique-string';
+import createHttpError from 'http-errors';
+import Articles from '../articles';
+
+export default (articles: Articles): Middleware => (
+  async ({ request, response }: Context, next: Next): Promise<void> => {
+    const article = request.body as JsonLdObj;
```
thewilkybarkid: Really we'll want the context to know that the body is JSON-LD. We can follow this up.

thewilkybarkid (approved):

PR #45, opened 2019-11-21
Use recursive Make idiomatically

giorgiosironi:
Prerequisites are evaluated in no specified order in parallel make, and support up to a certain number of simultaneous jobs that is specified globally.


Makefile
```diff
@@ -76,11 +76,9 @@ run:
...
-dev: ## Build and runs the container for development
-	make --jobs=4 install build stop
+dev: install build stop ## Build and runs the container for development
 	make run
 
 prod: export TARGET = prod
-prod: ## Builds and runs the container for production
-	make --jobs=2 build stop
+prod: build stop ## Builds and runs the container for production
 	make run
```
giorgiosironi: could move things around with intermediate targets to ensure `run` is a prerequisite as well, but seems harmless at the moment being a single command.

Makefile
```diff
@@ -2,6 +2,7 @@
 .PHONY: help install gitmodules build start stop wait-healthy sh exec logs watch lint fix test run dev prod
 
 SHELL = /usr/bin/env bash
+MAKEFLAGS += --jobs=4
```
giorgiosironi, line 5: this is an enforcement rather than a default. Haven't found a sane way to allow override with `--jobs` yet.

thewilkybarkid: This puts it at the end? If you can put it at beginning, the last value is used (ie this becomes a default).

Makefile
thewilkybarkid: Does still end up with `make[1]: warning: -jN forced in submake: disabling jobserver mode.`

giorgiosironi: It could be because this is using `make` rather than `$(MAKE)`; but do we want to go with the recursive make route or completely through prerequisites?

giorgiosironi:
While working on this discovered: https://github.com/hypermedia-app/generic.hypermedia.app/issues/46

giorgiosironi:
I think the build is running `start` and `wait-healthy` in parallel as the result of the new `MAKEFLAGS`.

thewilkybarkid:
Guess `--jobs=1` would have to be set on the outside, which is a bit annoying.

(Not with `MAKEFLAGS += --jobs=4` appending though.)

giorgiosironi:
An abundance of phony targets seems to be the problem, looking into the concepts of targets making an object (usually files) to see if it simplifies the picture.

giorgiosironi:
Committed two different possible ways to solve involuntary parallelism:

1. making dependencies like `start` before `wait-healthy` explicit
2. multiple `make` executions (as parallelism seems to override the documented order of targets in invocations like `make first second`)

Makefile
giorgiosironi: Can do this with
```
MAKEFLAGS ::= "-j 4 $(MAKEFLAGS)" 
```
but `MAKEFLAGS` is a magic variable so it gets rewritten to `-j --jobserver-fds=3,4` which I believe is what should be passed to recursive `$(MAKE)` invocations.

giorgiosironi:
Build failed (twice?) on:
```
npm ERR! 503 No backends available - GET https://registry.npmjs.org/@rdfjs/sink/-/sink-1.0.3.tgz
```

Makefile
giorgiosironi: So current status of this PR is to enforce the parallelism, but making sure that targets are never in a race condition. This should cater to maximum performance, but the number of total jobs is not customizable (we could use load average as the limit rather than jobs count, but the builds are not so long that it would be a useful metric).

.travis/analyse.sh
```diff
@@ -2,7 +2,7 @@
 set -e
 
 function finish() {
-  make logs stop
+  make logs && make stop
```
thewilkybarkid: `make logs; make stop`?

giorgiosironi: Yes, safer to always stop even in case of `docker-compose logs` failing.

thewilkybarkid (commented):
Not convinced the trade off is worth it. Feels like using `$(MAKE)` with `--jobs` and `$(MAKEFLAGS)` gives a better result.

giorgiosironi:
> using $(MAKE) with --jobs and $(MAKEFLAGS)

To my understanding this makes the job server available to sub-`make`s and achieves the parallelism. It is a special case that the Makefile being used is the main one rather than a sub-Makefile from another folder. Will change towards it to check this vision.

The larger point to me is that Makefiles containing only phony tasks are non-idiomatic as it's supposed to produce files as output rather than performing generic operations :sweat_smile:

giorgiosironi:
Updated not to introduce prerequisites but rather using recursive $(MAKE) which is stated to cover for the generated `MAKEFLAGS`. Can't use `--jobs` with that as the `MAKEFLAGS` pass down a centralized job server specifying a total number of jobs (or load).

thewilkybarkid:
I get what this is trying to do (use Make more 'correctly'), but don't see it as an improvement (and you now get a warning).

giorgiosironi:
I can't find warnings while executing this branch, but it is confusing different aspects like concurrency of targets and level of parallelism. I compared it with `master` and in there `make dev` was not working for me, so I opened https://github.com/libero/article-store/pull/83 instead.

PR #46, opened 2019-11-22
Fix node identifiers in the collection manages block

thewilkybarkid:
Currently they are plain strings rather than identifiers.


thewilkybarkid:
Looks like this would be caught by https://github.com/hypermedia-app/hydra-validator/issues/2.

giorgiosironi (approved):

PR #47, opened 2019-11-22
Retrieve articles from repository

nlisgo:
This introduces a retrieval process for the article listing to replace the empty hardcoded listing.


src/nodes.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export interface Nodes {
+    all(): Iterable<JsonLdObj>;
+
+    add(node: JsonLdObj): Promise<void>;
+
+    get(id: Iri): Promise<JsonLdObj>;
+
+    has(id: Iri): Promise<boolean>;
+}
```
thewilkybarkid: Suggest we put this in its own module, and try something like:

```typescript
export default interface Articles extends AsyncIterable<JsonLdObj> {
  count(): Promise<number>;

  get(id: Iri): Promise<JsonLdObj>;

  set(article: JsonLdObj): Promise<void>;

  delete(id: Iri): Promise<void>;

  has(id: Iri): Promise<boolean>;
};
```

test/routes/article-list.test.ts
```diff
@@ -1,13 +1,20 @@
...
 const makeRequest = async (next?: Next): Promise<Response> => {
   const context = createContext();
-
-  return runMiddleware(articleList(context.router), context, next);
+  const articles = {
+    all: (): Iterable<JsonLdObj> => [],
+    add: jest.fn(),
+    get: jest.fn(),
+    has: jest.fn(),
+  };
```
thewilkybarkid: We should create a proper `NullArticles` implementation which is empty (similar to https://github.com/libero/content-api-bundle/blob/master/src/Adapter/NullItems.php).

src/nodes.ts
```diff
@@ -0,0 +1,31 @@
+import { Iri, JsonLdObj } from 'jsonld/jsonld-spec';
+
+export interface Nodes {
+    all(): Iterable<JsonLdObj>;
+
+    add(node: JsonLdObj): Promise<void>;
```
stephenwf: Is there a narrower type for articles yet? Or is this intended to be more generic? Could do:
```typescript
export interface Nodes<T extends JsonLdObj = JsonLdObj> {
    // ...
    add(node: T): Promise<void>;
    // ...
}
```

if it is generic, to get typings when using it

thewilkybarkid: `@types/jsonld` doesn't provide much (`type JsonLdObj = object`), though could extend it to require `@id` and make `@type` contain `http://schema.org/Article`?

stephenwf: I wonder if https://github.com/google/schema-dts might be useful at some point. If it can convert a JSON-LD context (well, a list of triples) to typescript types.

thewilkybarkid: Just had a quick play with a less strict version of that:

```typescript
interface Article extends JsonLdObj {
  '@id': Iri;
  '@type': 'http://schema.org/Article' | Array<'http://schema.org/Article' | Iri>;
}
```

Doesn't really work as using the `rdf-namespaces` library is incompatible with the string literals. Different JSON-LD structures will make it hard to deal with too. Suggest the implementation itself validates instead.

src/routes/article-list.ts
```diff
@@ -2,9 +2,14 @@ import Router from '@koa/router';
 import { Context, Middleware, Next } from 'koa';
 import { hydra, rdf, schema } from 'rdf-namespaces';
 import Routes from './index';
+import { Nodes } from '../nodes';
 
-export default (router: Router): Middleware => (
+export default (articles: Nodes, router: Router): Middleware => (
   async ({ request, response }: Context, next: Next): Promise<void> => {
+    let list = [];
+    for await (const article of articles) {
```
nlisgo: this is currently causing an error:

```
FATAL ERROR: invalid array length Allocation failed - JavaScript heap out of memory
```

src/nodes.ts
```diff
@@ -1,30 +1,51 @@
...
-      return Object.values(this.nodes);
+    private i = 0;
+
+    [Symbol.asyncIterator](): AsyncIterator<JsonLdObj> {
+      const article = this.get(Object.keys(this.nodes)[this.i]);
+      this.i += 1;
+      return {
+        next(): Promise<IteratorResult<JsonLdObj>> {
+          return Promise.resolve({
+            done: false,
```
stephenwf: I think you need to set `done` to true to avoid the memory issue. Maybe something like:
```
done: !article,
```
would work?

src/nodes.ts
```diff
@@ -0,0 +1,53 @@
...
+    has(id: Iri): Promise<boolean>;
+}
+
+export default class InMemoryNodes implements Nodes {
+    private nodes: { [key: string]: JsonLdObj } = {};
+
+    private i = 0;
+
+    [Symbol.asyncIterator](): AsyncIterator<JsonLdObj> {
+      const isDone = Object.keys(this.nodes)[this.i] === undefined;
```
nlisgo: This now works when no nodes are set but we still experience the fatal error when we set any nodes.

src/nodes.ts
thewilkybarkid: Turned into https://github.com/libero/publisher/issues/346 (doesn't add a type though).

src/nodes.ts
```diff
@@ -15,14 +15,14 @@ export interface Nodes extends AsyncIterable<JsonLdObj> {
...
     private nodes: { [key: string]: JsonLdObj } = {};
 
-    private i = 0;
-
     [Symbol.asyncIterator](): AsyncIterator<JsonLdObj> {
-      const isDone = Object.keys(this.nodes)[this.i] === undefined;
-      const article = !isDone ? this.nodes[Object.keys(this.nodes)[this.i]] : undefined;
-      this.i += 1;
+      const nodes = this.nodes;
+      let i = 0;
```
Cooryd: 👍

src/adaptors/articles.ts
```diff
@@ -0,0 +1,13 @@
...
+    count(): Promise<number>;
+
+    get(id: Iri): Promise<JsonLdObj>;
+
+    set(article: JsonLdObj): Promise<void>;
+
+    delete(id: Iri): Promise<void>;
+
+    has(id: Iri): Promise<boolean>;
+}
```
nlisgo: I'm getting an eslint error here:

```
/app/src/adaptors/articles.ts
  13:2  error  Missing semicolon  semi
```

Is this possibly a bug with the linter?

src/routes/article-list.ts
```diff
@@ -2,9 +2,14 @@ import Router from '@koa/router';
 import { Context, Middleware, Next } from 'koa';
 import { hydra, rdf, schema } from 'rdf-namespaces';
 import Routes from './index';
+import Articles from '../adaptors/articles';
 
-export default (router: Router): Middleware => (
+export default (articles: Articles, router: Router): Middleware => (
   async ({ request, response }: Context, next: Next): Promise<void> => {
+    const list = [];
+    for await (const article of articles) {
```
nlisgo: With the current lint rules this triggers an error:

```
/app/src/routes/article-list.ts
  10:5  error  iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations  no-restricted-syntax
```

Should we consider adjusting our rules to allow for loops?

src/adaptors/null-articles.ts
```diff
@@ -0,0 +1,35 @@
+import { JsonLdObj, Iri } from 'jsonld/jsonld-spec';
+import Articles from './articles';
+
+export default class NullArticles implements Articles {
```
nlisgo: Many lint errors are introduced by this class:

```
/app/src/adaptors/null-articles.ts
   5:20  error    Expected 'this' to be used by class method                 class-methods-use-this
  16:14  error    Expected 'this' to be used by class async method 'count'   class-methods-use-this
  20:12  error    Expected 'this' to be used by class async method 'get'     class-methods-use-this
  24:12  error    Expected 'this' to be used by class async method 'set'     class-methods-use-this
  24:13  warning  'node' is defined but never used                           @typescript-eslint/no-unused-vars
  28:15  error    Expected 'this' to be used by class async method 'delete'  class-methods-use-this
  28:16  warning  'id' is defined but never used                             @typescript-eslint/no-unused-vars
  28:40  error    Unexpected empty async method 'delete'                     @typescript-eslint/no-empty-function
  32:12  error    Expected 'this' to be used by class async method 'has'     class-methods-use-this
  32:13  warning  'id' is defined but never used                             @typescript-eslint/no-unused-vars
```

src/nodes.ts
nlisgo: I'm resolving as this will be resolved by the linked issue.

test/routes/article-list.test.ts
nlisgo: I'm resolving as we have a `NullArticles` adaptor now.

src/nodes.ts
nlisgo: `Articles` now extends `Iterable` we can implement `AsyncIterable` as needed later.

src/adaptors/null-articles.ts
```diff
@@ -0,0 +1,35 @@
...
+  async get(id: Iri): Promise<JsonLdObj> {
+    throw new RangeError(`An article with the ID ${id} could not be found`);
+  }
+
+  async set(node: JsonLdObj): Promise<void> {
+    throw new Error('Unable to add an article');
+  }
+
+  async delete(id: Iri): Promise<void> {
+
```
thewilkybarkid: In this situation, I would have a `Do nothing` comment to make it explicit.

src/adaptors/null-articles.ts
thewilkybarkid: A lot of these don't sound right, but maybe could just use `InMemoryArticles` and don't add anything to it for now.

src/router.ts
```diff
@@ -3,12 +3,15 @@ import Routes from './routes';
 import apiDocumentation from './routes/api-documentation';
 import articleList from './routes/article-list';
 import entryPoint from './routes/entry-point';
+import InMemoryArticles from './adaptors/in-memory-articles';
+import Articles from './adaptors/articles';
 
 export default (): Router => {
   const router = new Router();
+  const articles: Articles = new InMemoryArticles();
```
thewilkybarkid: Create this at the top level and pass it in.

See also https://github.com/libero/article-store/pull/60.

src/adaptors/in-memory-articles.ts
```diff
@@ -0,0 +1,46 @@
...
+          const isDone = iri === undefined;
+          const article = !isDone ? nodes[iri] : undefined;
+          i += 1;
+          return {
+            done: isDone,
+            value: article,
+          };
+        },
+      };
+    }
```
thewilkybarkid: Should be able to simplify this with a generator, so something like

```typescript
* [Symbol.iterator](): Iterator<JsonLdObj> {
  yield* Object.values<JsonLdObj>(this.nodes);
}
```

src/adaptors/articles.ts
```diff
@@ -0,0 +1,13 @@
+import { Iri, JsonLdObj } from 'jsonld/jsonld-spec';
+
+export default interface Articles extends Iterable<JsonLdObj> {
```
thewilkybarkid: Following https://github.com/libero/publisher/issues/346, this should probably be a [`Store`](https://rdf.js.org/stream-spec/#store-interface)...

src/adaptors/null-articles.ts
nlisgo: Resolving as we have removed the NullArticles class for now.

src/adaptors/null-articles.ts
nlisgo: Resolving as we have removed the NullArticles class for now.

src/adaptors/articles.ts
```diff
@@ -0,0 +1,13 @@
+import { Iri, JsonLdObj } from 'jsonld/jsonld-spec';
+
+export interface Articles extends Iterable<JsonLdObj> {
```
nlisgo: Adding default here introduces the lint error: 

```
/app/src/adaptors/articles.ts
  13:2  error  Missing semicolon  semi
```

src/adaptors/articles.ts
giorgiosironi: My 2 cents would be that interfaces fall into the same realm as classes and functions (rather than expressions such as `{ ... }`) so they shouldn't require a semicolon. But the [rules I know](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export) apply to JavaScript rather than TypeScript, so unclear.

src/adaptors/in-memory-articles.ts
```diff
@@ -0,0 +1,33 @@
+import { JsonLdObj, Iri } from 'jsonld/jsonld-spec';
+import { Articles } from './articles';
+
+export default class InMemoryArticles implements Articles {
+    private nodes: { [key: string]: JsonLdObj } = {};
+
+    * [Symbol.iterator](): Iterator<JsonLdObj> {
```
giorgiosironi: That C feeling (I know, [not a pointer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/yield*))

src/adaptors/in-memory-articles.ts
```diff
@@ -0,0 +1,33 @@
...
+    async set(node: JsonLdObj): Promise<void> {
+      this.nodes[node['@id']] = node;
+    }
+
+    async delete(id: Iri): Promise<void> {
+      delete this.nodes[id];
+    }
+
+    async has(id: Iri): Promise<boolean> {
+      return id in this.nodes;
```
giorgiosironi: I guess the result is automatically boxed into a `Promise`

src/adaptors/in-memory-articles.ts
```diff
@@ -0,0 +1,33 @@
+import { JsonLdObj, Iri } from 'jsonld/jsonld-spec';
+import { Articles } from './articles';
+
+export default class InMemoryArticles implements Articles {
```
giorgiosironi, line 6: might be worth noting somewhere that this is not just an in-memory implementation but also a synchronous one

src/adaptors/in-memory-articles.ts
```diff
@@ -4,21 +4,8 @@ import Articles from './articles';
...
-          const article = !isDone ? nodes[iri] : undefined;
-          i += 1;
-          return {
-            done: isDone,
-            value: article,
-          };
-        },
-      };
+    * [Symbol.iterator](): Iterator<JsonLdObj> {
+      yield* Object.values<JsonLdObj>(this.nodes);
```
Cooryd: Great refactor 👍

src/routes/article-list.ts
```diff
@@ -2,8 +2,9 @@ import Router from '@koa/router';
 import { Context, Middleware, Next } from 'koa';
 import { hydra, rdf, schema } from 'rdf-namespaces';
 import Routes from './index';
+import { Articles } from '../adaptors/articles';
```
thewilkybarkid: The interface shouldn't be under `adaptors`.

test/routes/article-list.test.ts
```diff
@@ -3,11 +3,13 @@ import { Next, Response } from 'koa';
...
 
 const makeRequest = async (next?: Next): Promise<Response> => {
   const context = createContext();
+  const articles = new InMemoryArticles();
 
-  return runMiddleware(articleList(context.router), context, next);
+  return runMiddleware(articleList(articles, context.router), context, next);
 };
 
 describe('article list', (): void => {
```
thewilkybarkid, line 18: We should expand the tests to cover the empty list, but also a list with articles.

src/adaptors/in-memory-articles.ts
thewilkybarkid: Need tests.

src/adaptors/in-memory-articles.ts
thewilkybarkid: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Return_value

src/adaptors/articles.ts
```diff
@@ -0,0 +1,13 @@
...
+    count(): Promise<number>;
```
thewilkybarkid: Feels a bit like this should be `readonly length: Promise<number>` instead.

stephenwf: That might be confusing mixing with the `Array { length: number }` property.

thewilkybarkid: Was thinking about them in the same way, but the promise is different...

src/adaptors/in-memory-articles.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    async count(): Promise<number> {
+      return Object.values(this.nodes).length;
+    }
+
+    async get(id: Iri): Promise<JsonLdObj> {
+      if (this.has(id)) {
+        return this.nodes[id];
+      }
+      throw new RangeError(`An item with the ID ${id} could not be found`);
```
thewilkybarkid: Suggest we have an `ArticleNotFound` error.

thewilkybarkid: I'd prefer reversing this, generally throwing an error shouldn't be the default path (better to fail quickly).

src/adaptors/articles.ts
thewilkybarkid:
```typescript
interface Articles {}

export default Articles;
```

works, but then runs into https://github.com/typescript-eslint/typescript-eslint/issues/342. Let's disable `no-undef` (in a separate PR).

src/adaptors/articles.ts
```diff
@@ -0,0 +1,13 @@
...
+
+    get(id: Iri): Promise<JsonLdObj>;
+
+    set(article: JsonLdObj): Promise<void>;
```
thewilkybarkid: Should probably call this `add`, since it's not (visibly) assigning the object to something.

Question about if it's clear that it `update`s too...

thewilkybarkid: Though there's a question here about the identifier. Currently it's the full IRI/URI, and is using the scheme/host/port at runtime https://github.com/libero/article-store/blob/10f133673ef1abe8cc9597a535b6128adf087394/src/routes/article-list.ts#L10

I've never been entirely happy with this (what happens if it changes?), so maybe it is time to introduce a base URI setting that it used everywhere instead (and is hard/impossible to change). Means if the app doesn't have to care if it's behind a proxy.

src/adaptors/articles.ts
thewilkybarkid: https://github.com/libero/article-store/pull/47#discussion_r352084039

src/adaptors/articles.ts
stephenwf: In case it helps, default exports do have downsides:
https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html

src/adaptors/articles.ts
thewilkybarkid: @stephenwf @tpluscode Any advice?

stephenwf: Just so I understand:
* A document list is served with a generated `@id` property based on the origin
* POST-ed documents will have an `@id` in the body of their document
* POST-ed documents will hit an endpoint `/articles/{identifier}`
* Documents will be saved with an `@id` based on [ ... ]
* Documents will be served with an `@id` based on [ ... ]

Seems like an identified can't really be tied to a base domain, or changes to that domain (like proxy or environments) will be really tough. Will we save a non URI based identifier that is more portable that can be used to generate `@id` identifiers using some configuration?

In general though, problems start to arise when you can't rely on the router in order to dereference an `@id`, so for any embedded documents or adding relations and on the client side they may have to know how to construct a URI from an `@id` they are given.

If you do store an identifier that is not tied to a domain/protocol/host, I would suggest that you also include it in the body of the document under something like `dcterms:identifier` or similar.

test/routes/article-list.test.ts
```diff
@@ -3,23 +3,65 @@ import { Next, Response } from 'koa';
...
+      articles.add({
+        '@type': 'http://schema.org/Article',
+        '@id': '/articles/24231',
+        'http://schema.org/name': 'The age of Homo naledi and associated sediments in the Rising Star Cave, South Africa',
+      }),
+    ]);
+  })();
+
+  return runMiddleware(articleList(articles, context.router), context, next);
 };
```
thewilkybarkid: Rather than have two, I'd make `articles` an optional argument (default to an empty `InMemoryArticles`), then add in the items in the test itself.

test/routes/article-list.test.ts
```diff
@@ -3,23 +3,65 @@ import { Next, Response } from 'koa';
...
+        '@id': '/articles/09560',
+        'http://schema.org/name': 'Homo naledi, a new species of the genus Homo from the Dinaledi Chamber, South Africa',
+      }),
+      articles.add({
+        '@type': 'http://schema.org/Article',
+        '@id': '/articles/24231',
+        'http://schema.org/name': 'The age of Homo naledi and associated sediments in the Rising Star Cave, South Africa',
+      }),
+    ]);
+  })();
```
thewilkybarkid: Function is `async` so can just `await` the `add`s individually/together?

src/adaptors/articles.ts
thewilkybarkid: Hmm, eslint will currently complain if trying to do this on a class. Favour sticking with defaults for now (for consistency), but one to have a think about.

src/routes/article-list.ts
nlisgo: I'm moving it to the `/src` folder or should we organise it somewhere else?

test/routes/article-list.test.ts
```diff
@@ -3,23 +3,65 @@ import { Next, Response } from 'koa';
...
+
+const makeRequestList = async (next?: Next): Promise<Response> => {
+  const context = createContext();
+  const articles = new InMemoryArticles();
+
+  (async (): Promise<void> => {
+    await Promise.all([
+      articles.add({
+        '@type': 'http://schema.org/Article',
+        '@id': '/articles/09560',
```
thewilkybarkid: Since these aren't dereferenceable yet, we should use a [blank node](https://www.w3.org/TR/rdf11-concepts/#h3_section-blank-nodes), so like `'@id': '_:12345'`.

To make the Hydra console work, we need to let Hydra know about the class too. So the API documentation will need:

```typescript
{
  '@id': schema.Article,
  '@type': hydra.Class,
  [hydra.title]: { '@value': 'Article', '@language': 'en' },
  [hydra.supportedProperty]: [
    {
      '@type': hydra.SupportedProperty,
      [hydra.title]: { '@value': 'Name', '@language': 'en' },
      [hydra.property]: {
        '@id': schema.name,
        '@type': rdf.Property,
      },
      [hydra.required]: true,
      [hydra.readable]: true,
      [hydra.writeable]: true,
    },
  ],
},
```

(Try then adding an article manually in `app.ts` and see it appear.)

src/adaptors/articles.ts
giorgiosironi: > Should probably call this add, since it's not (visibly) assigning the object to something.

`add` and `remove` are collection (Repository pattern) terms, they make sure an object is in the collection or that it isn't with the intent that only the collection is persisted while objects are transient. `add` then doesn't cover update scenarios because the object is already in the collection, and it is directly updated after a `find`.

A Unit Of Work separately keeps track of which in-memory objects have been modified and should be reflected into persistence. This works seamlessly also for this in-memory implementation, as you retrieve the object and modify it while the map inside here still has a reference to it.

thewilkybarkid: We'll need to generate a local UUID or something for each article, but externally the identifier is the URI itself (so isn't something the user will construct). We will probably need a way to be able to find an article by a user-defined identifier though (eg DOI).

Wonder how feasible it is to internally use a different format, then convert it to the public URI on the way out. (Like locally just referencing `/articles/12345`, then making it `http://example.com/prefix/articles/12345`.)

giorgiosironi: The `Repositories` chapter of https://www.oreilly.com/library/view/implementing-domain-driven-design/9780133039900/ talks about Persistence-oriented Repositories (as opposed to Collection-oriented) and in that case a universal (upsert) `save()` method is used along with a `remove()` one.

thewilkybarkid: Going to delay the ID question as we can just use blank nodes for now.

src/adaptors/in-memory-articles.ts
stephenwf: Not sure if it factors in, but code that uses `Articles` interface expecting async should be able to use `await` on non promise values (as in not have to check if it is async or not, it just works). Only diff is `await` pushes a micro-task, meaning you can get some slightly unexpected ordering. Demo for the curious: https://codesandbox.io/s/bold-dew-u3iz9?expanddevtools=1&fontsize=14&hidenavigation=1&theme=dark (async/await is strange)

src/adaptors/in-memory-articles.ts
```diff
@@ -0,0 +1,39 @@
...
+import Articles from '../articles';
+import ArticleNotFound from '../errors/article-not-found';
+import InvalidId from '../errors/invalid-id';
+
+export default class InMemoryArticles implements Articles {
+  private nodes: { [key: string]: JsonLdObj } = {};
+
+  async add(node: JsonLdObj): Promise<void> {
+    if (node['@id'] === undefined) {
+      throw new InvalidId();
```
thewilkybarkid: `ArticleHasNoId` or something? (The ID itself isn't invalid as it doesn't exist.)

thewilkybarkid (approved):

PR #48, opened 2019-11-22
Fix collection title property

thewilkybarkid:
Uses `hydra:title` not `rdfs:label` https://github.com/libero/article-store/blob/ccc19172f83cf7a22ec75353aadbb74ac25cdb72/src/routes/article-list.ts#L14


thewilkybarkid:
@tpluscode Could this be a Hydra Validator check? (When a `hydra:required` property is missing.)

PR #49, opened 2019-11-22
Define hydra:collection as a supported property on the entry point

thewilkybarkid:
Adds `hydra:collection` to the API documentation to clients are aware that it is a `hydra:Link` without having to separately know it.


thewilkybarkid:
Left in WIP as the Hydra console displays it twice:

Screenshot 2019-11-22 at 16 16 31

Fiddling with titles, the top one comes from the API documentation and the other from the entry point resource itself.

thewilkybarkid:
Debugged it, comments in https://github.com/wikibus/Alcaeus/issues/142#issuecomment-558091335.

PR #56, opened 2019-11-25
Allow Dependabot to auto-merge non-major updates

thewilkybarkid:
Dependabot will be opening quite a few PRs, this allows it to auto-merge minor and patch updates. 

There's always a chance that it will introduce problems even if the status checks all pass, but a human wouldn't have found it anyway...

Refs https://github.com/libero/article-store/pull/51#issuecomment-558090886.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -7,6 +7,10 @@ update_configs:
     commit_message:
       prefix: chore
       include_scope: true
+    automerged_updates:
+      - match:
+          dependency_type: all
+          update_type: 'semver:minor'
     default_labels:
       - ':recycle: Chore'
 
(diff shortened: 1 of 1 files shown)


.dependabot/config.yml
```diff
@@ -25,5 +33,9 @@ update_configs:
     commit_message:
       prefix: chore
       include_scope: true
+    automerged_updates:
+      - match:
+          dependency_type: all
+          update_type: 'semver:minor'
```
giorgiosironi: what does this mean here? Our submodule only has commits, no tags

giorgiosironi: https://github.com/libero/scripts/ should be one of the repositories where we never make breaking changes given how entrenched it becomes with builds and deployments, so could probably be always updated then.

thewilkybarkid: Never say never.  😛

While we could tag, maybe let's just not auto merge these for now.

giorgiosironi (approved):

PR #60, opened 2019-11-26
Use the Koa context rather than handling dependencies separately

thewilkybarkid:
Current the router is being passed around separately, rather than using the Koa context. This adopts the pattern used by the framework (and is already implemented in the `@koa/router` library).

I've kept the factory pattern in place, not sure whether to remove it where possible.


src/index.ts
```diff
@@ -1,3 +1,24 @@
...
+
+app.use(logger());
+app.use(cors({
+  exposeHeaders: ['Link'],
+}));
+app.use(apiDocumentation());
+app.use(errorHandler());
+app.use(routing(app.context.router));
+
+export default app.listen(8080);
```
thewilkybarkid: This change should be extracted.

giorgiosironi: https://github.com/libero/article-store/pull/80 approved and will make this PR leaner.

giorgiosironi: Actually doesn't look like the same thing, but will surely conflict?

thewilkybarkid: Just pushed https://github.com/libero/article-store/pull/60/commits/6be8217544419a4127648caf4f6abda4bc96a82a which updates this PR to include #80.

thewilkybarkid:
Really wanted `types.ts` to be in `app.ts`, but end up with circular references.

src/types.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export type AppState = DefaultState;
+
+export type AppContext = RouterContext<AppState, {
+  articles: Articles;
+}>;
+
+export type AppMiddleware = Middleware<AppState, AppContext>;
+
+export type App = Koa<AppState, AppContext>;
```
thewilkybarkid: Did want these inside `app.ts`, but end up with circular references otherwise.

stephenwf: Is `App` required? It should be inferred correctly from the default export returning in `app.ts`

thewilkybarkid: Looking at whether it can be reduced to just `AppContext`, as that's the important one.

thewilkybarkid: Nope. The separate router is the problem...

thewilkybarkid: In fact, even moving `createRouter` doesn't help (since the routes themselves are now circular).

thewilkybarkid: Might have to inject the router into the app factory? Shame, as the `apiDocumentation` middleware relies on a specific route. Might have to inject that too?

(Feels like testing will be a bit weirder.)

stephenwf: You could avoid that by making the route generic? 

```typescript
function createRouter<State, Context>() {
  const router = new Router<State, Context>();

  router.get( 
     //... 
  );
}
```
Although might need to narrow the types on the generics

thewilkybarkid: Tried that, not assignable in https://github.com/libero/article-store/blob/1acc2dadafdd154f9b6f5f61fb9545524da29be3/src/app.ts#L17 😞

thewilkybarkid: Thinking about making the path an argument on `apiDocumentation`, so it's decoupled from the router.

src/routes/article-list.ts
```diff
@@ -1,11 +1,12 @@
...
 import { hydra, rdf, schema } from 'rdf-namespaces';
-import Articles from '../articles';
+import { AppContext, AppMiddleware } from '../types';
 import Routes from './index';
 
-export default (articles: Articles, router: Router): Middleware => (
-  async ({ request, response }: Context, next: Next): Promise<void> => {
+export default (): AppMiddleware => (
+  async ({
+    articles, request, response, router,
```
giorgiosironi, line 8: it is a refactoring from Dependency Injection to a Service Locator (the `AppContext`), but at least this unpacking makes the dependencies clear

src/types.ts
thewilkybarkid: https://github.com/libero/article-store/pull/60/commits/bf78b85d521ce96af83994c6611e7a783b795a60, but needs extracting

stephenwf: Right, I see. Pulled it down to check and yeh it's like the app and router are pretty must one and the same conceptually. I definitely see the benefit in splitting out the routes though.

In this case though, it does make sense to have some types grouped together as a common concept that they are all a part of (the "app").

thewilkybarkid: Push https://github.com/libero/article-store/pull/60/commits/14962894eb80dce81cceb5f12cba8fc23c8a479e which is now possible. Doesn't feel great though.

src/app.ts
```diff
@@ -1,16 +1,30 @@
...
 import errorHandler from './middleware/error-handler';
 import routing from './middleware/routing';
-import createRouter from './router';
 
-export default (articles: Articles): Koa => {
-  const app = new Koa();
-  const router = createRouter(articles);
+export type AppState = DefaultState;
+
+export type AppContext = RouterContext<AppState, {
```
giorgiosironi, line 13: seems like a good alias

giorgiosironi: does it matter that types are inside a top level module like `app`? Though they are used in conjunction with this module so it's probably the right place even if requiring this brings in everything.

src/index.ts
```diff
@@ -1,8 +1,12 @@
...
 import createApp from './app';
+import createRouter from './router';
+import Routes from './routes';
 
 const articles = new InMemoryArticles();
+const router = createRouter();
+const apiDocumentationPath = router.url(Routes.ApiDocumentation);
 
-const app = createApp(articles);
+const app = createApp(articles, router, apiDocumentationPath);
```
giorgiosironi, line 10: doesn't look like it can scale to lots of dependencies?

test/router.test.ts
```diff
@@ -1,10 +1,9 @@
 import Layer from '@koa/router/lib/layer';
-import InMemoryArticles from '../src/adaptors/in-memory-articles';
 import createRouter from '../src/router';
 import Routes from '../src/routes';
 
 describe('router', (): void => {
-  const router = createRouter(new InMemoryArticles());
+  const router = createRouter();
```
giorgiosironi, line 6: sounds good that to create a router you don't need an `Articles` implementation :thinking:

test/routes/entry-point.test.ts
```diff
@@ -4,11 +4,9 @@ import entryPoint from '../../src/routes/entry-point';
 import runMiddleware from '../middleware';
 import createContext from '../context';
 
-const makeRequest = async (next?: Next): Promise<Response> => {
-  const context = createContext();
-
-  return runMiddleware(entryPoint(context.router), context, next);
-};
+const makeRequest = async (next?: Next): Promise<Response> => (
+  runMiddleware(entryPoint(), createContext(), next)
```
giorgiosironi, line 8: `makeRequest` doesn't return but can still `await` on it to get a response?

thewilkybarkid: Parentheses rather than curly brackets (line above) are a return.

giorgiosironi: ah, it's Ruby all over again :+1:

giorgiosironi (approved):
Long but mostly consistent across updating routes and tests

thewilkybarkid:
Will do https://github.com/libero/article-store/pull/60#discussion_r354792149 before merging this.

src/index.ts
thewilkybarkid: Could remove `apiDocumentationPath`, and assume the router has a route named `Routes.ApiDocumentation`...

PR #61, opened 2019-11-26
Only build the master branch

thewilkybarkid:
Pull requests should come from a fork, but Dependabot uses the origin repo. (And sometimes forks are accidentally not used, eg #45.) These are currently built twice (branch, and the pull request). This stops the branch itself being built.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: minimal
 
 services: docker
 
+branches:
+  only:
+    - master
+
 env:
   global:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,6 +2,10 @@ language: minimal
 
 services: docker
 
+branches:
+  only:
+    - master
```
giorgiosironi, line 7: guess we don't need something more elaborate like https://github.com/libero/browser/blob/master/.travis.yml#L35-L39

giorgiosironi (approved):

PR #64, opened 2019-11-28
Use RDF/JS rather than JSON-LD

thewilkybarkid:
Rather than using JSON-LD, this uses the [RDF JavaScript interfaces and libraries](https://rdf.js.org/) instead.


src/routes/api-documentation.ts
```diff
@@ -1,106 +1,107 @@
...
+      quad(entryPointGet, hydra('returns'), schema('EntryPoint')),
+      quad(schema('EntryPoint'), hydra('supportedProperty'), entryPointName),
+      quad(entryPointName, rdf('type'), hydra('SupportedProperty')),
+      quad(entryPointName, hydra('title'), literal('Name', 'en')),
+      quad(entryPointName, hydra('property'), schema('name')),
+      quad(schema('name'), rdf('type'), rdf('Property')),
+      quad(entryPointName, hydra('required'), toRdf(true)),
+      quad(entryPointName, hydra('readable'), toRdf(true)),
+      quad(entryPointName, hydra('writeable'), toRdf(false)),
+    );
```
thewilkybarkid, line 50: Clearly this isn't that easy to read.

Wondering if it could be a Turtle (or similar) string that gets parsed. Will leave for now though.

test/rdf.ts
```diff
@@ -0,0 +1,20 @@
...
+export const captureSource = (response: Response): Source => {
+  if (!isSource(response.body)) {
+    throw new TypeError('not a source');
+  }
+
+  return response.body;
+};
+
+export const toArray = async (stream: Stream): Promise<Array<Quad>> => (
+  streamToArray(stream as Stream & NodeJS.ReadableStream)
```
thewilkybarkid: Don't like `Stream & NodeJS.ReadableStream`.

src/routes/article-list.ts
```diff
@@ -1,28 +1,49 @@
...
+        carry.push(quad(articleList, hydra('member'), subject));
+
+        (await streamToArray(articles.match(null, null, null, subject)))
+          .forEach(({ predicate, object }: Quad): void => {
+            carry.push(quad(subject, predicate, object));
+          });
+
+        return carry;
+      }, Promise.resolve(quads),
+    );
```
thewilkybarkid, line 44: This is very messy. Think can use regex for the second match to reduce it to one call, but don't think it's possible to directly paginate on the first.

thewilkybarkid: Thinking we'll still want an `Articles`, which could return a `Stream` or a `Dataset`.

thewilkybarkid:
Going to recreate this PR using datasets. Too many changes to update with, to then be largely replaced.

PR #67, opened 2019-11-29
Adds schema:Article to the API documentation

thewilkybarkid:
The article list, although always empty, references `schema:Article`. This adds it to the API documentation.

Extracted from https://github.com/libero/article-store/pull/47.

### DIFF
--- a/src/routes/api-documentation.ts
+++ b/src/routes/api-documentation.ts
@@ -105,7 +105,7 @@ export default (router: Router): Middleware => (
           [hydra.supportedProperty]: [
             {
               '@type': hydra.SupportedProperty,
-              [hydra.title]: { '@value': 'Name', '@language': 'en' },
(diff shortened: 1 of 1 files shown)


src/routes/api-documentation.ts
```diff
@@ -98,6 +98,24 @@ export default (router: Router): Middleware => (
...
           ],
         },
+        {
+          '@id': schema.Article,
+          '@type': hydra.Class,
+          [hydra.title]: { '@value': 'Article', '@language': 'en' },
+          [hydra.supportedProperty]: [
+            {
+              '@type': hydra.SupportedProperty,
+              [hydra.title]: { '@value': 'Name', '@language': 'en' },
```
giorgiosironi: could `@value` be something else (like `Article title`), if we wanted it to be?

giorgiosironi (approved):

PR #71, opened 2019-12-02
Build on Github Actions

giorgiosironi:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,9 @@ name: CI
 
 on: [pull_request] # TODO: also `master`
 
+env:
+  IMAGE_TAG: github.sha
+
 jobs:
 
   build:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on: [pull_request] # TODO: also `master`
 
 jobs:
 
-  make:
+  build:
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+    - name: Smoke tests
+      run: |
+        set -e
+        TARGET=prod .travis/smoke-test.sh
+        TARGET=dev .travis/smoke-test.sh
+
+    - name: Lint
+      run: make lint
+
+    - name: Test
```
giorgiosironi, line 32: downstream steps are not executed if a previous one fails, at the moment

giorgiosironi:
Opened on the origin repository as forks seem to be poorly supported at the moment.

.github/workflows/ci.yml
thewilkybarkid: What's the cost of separate actions? Multiple Docker builds?

giorgiosironi: Seems multiple jobs, which start in isolated environments and hence Docker builds; unless the image can be shared as an artifact which is possible in principle but likely convoluted. Jobs enable parallelization as they can run on separate environments (actual VMs?)

We may also runn stuff in parallel at the microlevel (`make`, `xargs`) but they will be seen as a single step so doesn't make the failures clear. In general there are some dependencies between these steps, if the smoke tests do not run also `analyse.sh` likely won't (besides dependencies towards `make build` which previously was separated in Travis CI's `install` section).

giorgiosironi:
Possibly the missing event `push` ~`pull_request`~ is causing the fork problems. Will close this and keep working from the fork.

PR #73, opened 2019-12-02
Build with GitHub actions (PR from a fork)

giorgiosironi:


giorgiosironi:
Two checks appear, but it's alright as one was already existing from https://github.com/libero/article-store/pull/71. The results of the two PRs merge with each other but this is a super corner case.

PR #74, opened 2019-12-02
Build with GitHub actions

giorgiosironi:

Equivalent of the Travis CI one. Would remove the Travis CI build as a follow-up rather than atomically switch.


giorgiosironi:
Double check appears on https://github.com/libero/article-store/pull/74/commits/f268afbc299956a76a1e9a1f18dd604a17b0324a because the PR was reopened.

giorgiosironi:
Current serial version clocks in at ~3m11s (`11:35:34` to `11:38:45`).
https://github.com/giorgiosironi/article-store/pull/2/checks takes even more time, ~3m35s (`11:29:01` to `11:32:36`).

Makefile
```diff
@@ -34,7 +34,7 @@ start: ## Start the containers
 stop: ## Stop the containers
 	${DOCKER_COMPOSE} down
 
-wait-healthy: ## Wait for the containers to be healthy
+wait-healthy: gitmodules ## Wait for the containers to be healthy
```
thewilkybarkid: Isn't related?

Makefile
```diff
@@ -52,7 +52,7 @@ exec: ## Run a command on the app container
 		echo "No command provided"; \
 		exit 1; \
 	fi; \
-	${DOCKER_COMPOSE} exec app $(command)
+	${DOCKER_COMPOSE} exec -T app $(command)
```
thewilkybarkid: What happens when there is a TTY? Should be extracted to a separate PR too.

Makefile
giorgiosironi: Travis CI doesn't require this because it checks out Git submodules on its own, whereas the `actions/checkout` Github Action doesn't so needs this as a dependency

Makefile
giorgiosironi: Strange that removing didn't affect the shell Make target. Will experiment.

Makefile
thewilkybarkid: Oh hang on, script's in the submodule.

Makefile
giorgiosironi: The prompt and niceties like completion disappear when running `sh` with `-T` (which means remove the TTY), so will want to keep the command normal for interactive use outside of CI.

Makefile
giorgiosironi: Extract or leave?

Makefile
giorgiosironi: In https://github.com/libero/article-store/pull/77

Makefile
thewilkybarkid: Extract

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
+name: CI
+
+on: [push, pull_request]
```
thewilkybarkid: `push` has the same problem as https://github.com/libero/article-store/pull/61?

Makefile
thewilkybarkid: In fact, looks like it can do the same as Travis? https://github.com/actions/checkout/blob/50fbc622fc4ef5163becd7fab6573eac35f8462e/action.yml#L13-L14

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Build Docker images
+      run: |
+        set -e
```
thewilkybarkid: Not needed: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
...
+        TARGET=prod make build
+        TARGET=dev make build
+
+    - name: Lint
+      run: make lint
+      if: always()
```
thewilkybarkid: Meant? Looks like it will be forced to run even if the workflow is cancelled.

thewilkybarkid:
> Current serial version clocks in at ~3m11s (`11:35:34` to `11:38:45`).
> https://github.com/giorgiosironi/article-store/pull/2/checks takes even more time, ~3m35s (`11:29:01` to `11:32:36`).

Probably out of scope for this PR, but would be interested to know what happens when there's a `build` job, then all the others run in parallel (using something like https://stackoverflow.com/a/57877438).

thewilkybarkid:
Should add the badge to the README.

giorgiosironi:
> would be interested to know what happens when there's a build job, then all the others run in parallel

Isn't that what https://github.com/giorgiosironi/article-store/pull/2/checks did? One `build` plus 5 parallel jobs `lint`, `test`, `smoke-tests-prod`, `smoke-tests-dev`, `analyse`. Timestamps available with the hidden button `Show timestamp`.

.github/workflows/ci.yml
giorgiosironi: Only alternative I see to execute in spite of other steps failing would be `success() || failure()` which seems confusing?

thewilkybarkid:
> Isn't that what https://github.com/giorgiosironi/article-store/pull/2/checks did? One `build` plus 5 parallel jobs `lint`, `test`, `smoke-tests-prod`, `smoke-tests-dev`, `analyse`. Timestamps available with the hidden button `Show timestamp`.

Ah ok. Looks like that took ~3 minutes 30ish? Not that much slower, and the separate status checks are useful.

.github/workflows/ci.yml
giorgiosironi: Yes. https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#pull-request-event-pull_request states:
> Note: By default, a workflow only runs when a pull_request's activity type is `opened`, `synchronize`, or `reopened`. To trigger workflows for more activity types, use the types keyword.

This doesn't seem to cover new commits on an open pull request (nor do the other activity types listed there?). Will try.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,39 @@
+name: CI
+
+on: [pull_request]
```
giorgiosironi: https://github.com/libero/article-store/pull/74/checks?check_run_id=331381010 triggered so I'd leave it like this until it breaks.

thewilkybarkid: Looks like there's more granular options for creation etc: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#onevent_nametypes

giorgiosironi:
seems to be eagerly green despite this not having run on `master`.

giorgiosironi:
https://github.com/libero/article-store/pull/74/commits/3f33297405c533005badd767458ce02cf3f2edfb stopped building on PRs.

giorgiosironi:
> Looks like that took ~3 minutes 30ish? Not that much slower, and the separate status checks are useful.

If the project grows that could become more performant as it scales to larger amount of tests, and we don't pay for resources; at the moment it's 3x the amount of code however.

thewilkybarkid:
Wonder if it supports some of the fancier YAML syntax? 🤔

giorgiosironi:
Tried https://github.com/libero/article-store/pull/79 for parallelization but YAML anchors are not supported.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
+name: CI
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
giorgiosironi, line 5: still looking for some documentation about what `synchronize` is. It seems not to be just a merge of `master` into the PR's branch but also adding any commit. Will try removing to see what happens.

giorgiosironi:
No more improvements in the queue, ready for a new pass.

thewilkybarkid:
Going to borrow this PR to try something, will revert if it doesn't work.

thewilkybarkid:
Close to working, but lost the `dev` smoke test.

Feels a bit of a hack though. Tempted to say my preference is to have the separate jobs and live with the duplication. It's a bit annoying, but isn't _that_ bad, and gives us the most useful output. (Separate status checks with sensible names.)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        test:
+          - make lint
+          - make test
+          - .travis/smoke-test.sh
+          - .travis/analyse.sh
+        include:
```
giorgiosironi, line 62: reserved keyword in the middle of arbitrary arguments :face_with_head_bandage:

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+          - env: prod
```
giorgiosironi: smoke tests for `prod` have priority over `dev` ones

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+      - name: Download Docker image artifact
+        uses: actions/download-artifact@v1
+        with:
+          name: docker-image-${{ matrix.env || 'dev' }}
+
+      - name: Load Docker image
+        run: docker load -i docker-image-${{ matrix.env || 'dev' }}/article-store.tar
+
+      - name: Run test
+        run: ${{ matrix.test }}
```
giorgiosironi, line 82: a good thing about a standardized test job is that it prevents logic creeping into this not-actually-YAML file and pushes into scripts and Makefile targets, that can also be executed locally

.github/workflows/ci.yml
giorgiosironi: now runs both smoke tests

.github/workflows/ci.yml
```diff
@@ -0,0 +1,78 @@
...
+    strategy:
+      matrix:
+        test:
+          - make lint
+          - make test
+          - TARGET=dev .travis/smoke-test.sh
+          - TARGET=prod .travis/smoke-test.sh
+          - .travis/analyse.sh
+        include:
+          - env: prod
```
giorgiosironi: though this is `target` rather than `env`

giorgiosironi:
Current status:

- using a matrix to remove all duplication
- testing targets are independent
- their commit status checks are independent
- overhead of parallelism reduced via compression

We could question the name of the status checks to specify them in [branch protection](https://github.com/libero/article-store/settings/branch_protection_rules/11906254). But given how the infrastructure is managed I'd rather specify they all have to pass rather than listing them all there (assuming they will always appear as in progress or failed).

Build performance can follow reasonable patterns but will always be limited because the infrastructure is standardized. Parallelization here helps in the long run with respect to custom hardware.

giorgiosironi:
First hanging build? https://github.com/libero/article-store/runs/335171718

giorgiosironi:
With 3 jobs (part of the same matrix) blocked on the same step.

giorgiosironi:
`Cancel check suite` does turn off the yellow commit statuses.

giorgiosironi:
All jobs were blocked on
```
version-0.12.2: Pulling from tpluscode/hypermedia-app
```
Can only see this now, as the output was not accessible while they were running.

giorgiosironi:
No problems in pulling locally with:
```
docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.dev.yml pull console
```
```
$ docker image ls | grep hypermedia-app
tpluscode/hypermedia-app                                  version-0.12.2                             6f4fd57ea35f        11 days ago         125MB
generic-hypermedia-app                                    latest                                     6a02fdf92926        13 days ago         125MB
tpluscode/hypermedia-app                                  version-0.12.1                             d6a905c98767        2 weeks ago         125MB
tpluscode/hypermedia-app                                  alpha                                      16099832ee8d        3 weeks ago         124MB
```

giorgiosironi:
Probably a Docker Hub fluke.

thewilkybarkid (approved):

PR #77, opened 2019-12-03
Make analyse.sh non-interactive

giorgiosironi:
Extracted from https://github.com/libero/article-store/pull/74

`docker-compose exec` (without `-T`) is nice for interactive use, but is not friendly to automated systems running scripts such as Github Actions.

Typical failure:
```
docker-compose --file .docker/docker-compose.yml --file
.docker/docker-compose.dev.yml exec app npx hydra-validator analyse
http://localhost:8080/
the input device is not a TTY
```

### DIFF
--- a/.travis/analyse.sh
+++ b/.travis/analyse.sh
@@ -9,4 +9,4 @@ trap finish EXIT
 
 make start wait-healthy
 
-make exec command="npx hydra-validator analyse http://localhost:8080/"
+make docker-compose command="exec -T npx hydra-validator analyse http://localhost:8080/"
--- a/.travis/analyse.sh
+++ b/.travis/analyse.sh
@@ -9,4 +9,4 @@ trap finish EXIT
 
 make start wait-healthy
 
-make docker-compose command="exec -T npx hydra-validator analyse http://localhost:8080/"
+make docker-compose command="exec -T app npx hydra-validator analyse http://localhost:8080/"
--- a/.travis/analyse.sh
+++ b/.travis/analyse.sh
@@ -9,4 +9,4 @@ trap finish EXIT
 
 make start wait-healthy
 
-make docker-compose command="exec -T app npx hydra-validator analyse http://localhost:8080/"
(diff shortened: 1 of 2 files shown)


giorgiosironi:
Wait, how was this working in https://github.com/libero/article-store/pull/74 then?
```
docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.dev.yml exec -T npx hydra-validator analyse http://localhost:8080/
ERROR: No such service: npx
```

giorgiosironi:
Problem Exists Between Keyboard And Chair.

.travis/analyse.sh
```diff
@@ -9,4 +9,4 @@ trap finish EXIT
 
 make start wait-healthy
 
-make exec command="npx hydra-validator analyse http://localhost:8080/"
+make docker-compose command="exec -T app npx hydra-validator analyse http://localhost:8080/"
```
thewilkybarkid: Could make it an argument on Make's `exec`?

giorgiosironi: No preference, will have an ugly name too (will copy from `docker-compose exec --help`)

thewilkybarkid: Prefer to isolate Docker Compose to the Makefile entirely.

Makefile
```diff
@@ -47,12 +47,18 @@ wait-healthy: ## Wait for the containers to be healthy
 sh: ## Open a shell on the app container
 	make exec command="sh"
 
+tty=1
+ifeq ($(tty), 0)
+	options=-T
+else
+	options=
+endif
```
giorgiosironi: have to do this with a Make conditional rather than a bash one because it's proving difficult to interpolate an `options` variable inside the `docker-compose ...` command without creating a subshell. The `command` check could be dealt with in the same way, but this approach is executed when loading the Makefile rather than when executing particular targets.

giorgiosironi: These are also global. Will substitute with `$(if ...)`

thewilkybarkid (approved):

PR #80, opened 2019-12-04
Create the app using a factory

thewilkybarkid:
~Removes the separate factory.~

~Originally this was added as I thought Supertest wouldn't handle it, but it's fine. (It was immediately being bound to a port anyway.)~

Extracted from https://github.com/libero/article-store/pull/60#discussion_r353715319.


giorgiosironi:
The tests look like they have some problem in terminating the listening and allow Jest to exit:
```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

thewilkybarkid:
Ah, `--detectOpenHandles` shows it is the listen. (Can find locally, but Jest doesn't show the normal message. 🤔)

thewilkybarkid:
Hmm, going to have to rethink. Will need a separate factory and server.

thewilkybarkid:
Hmm, still not right. Not going to be able to avoid https://github.com/libero/article-store/blob/18830709c24f235e320c30a58d9faaa45abfba5e/src/app.ts#L5-L13 being defined alongside the `app` itself...

thewilkybarkid:
~Nope, still end up bit a big dependency loop in #60... Going to have to drop this type idea I think.~ False positive in PhpStorm, don't think its index had updated.

test/app.test.ts
```diff
@@ -1,10 +1,18 @@
...
+import createApp from '../src/app';
 
 const trim = (value: string): string => value.trim();
 const parseHeader = (header: string): Array<string> => header.split(',').map(trim);
 
+let app: Koa;
+
+beforeEach((): void => {
+  app = createApp(new InMemoryArticles());
+});
```
giorgiosironi, line 14: I understand no tear down is needed because this `app` instance doesn't listen on a port but is only used through in-process tests. Therefore it will just be garbage collected. The new instance of `InMemoryArticles` also ensures no state is carry over between tests.

giorgiosironi (approved):

PR #83, opened 2019-12-05
Let MAKEFLAGS be passed through the environment

giorgiosironi:
`make dev` is broken for me on `master`:
```
$ make dev
make --jobs=4 install build stop
make[1]: Entering directory '/home/giorgio/code/article-store'
make --jobs=2 w -j --jobserver-fds=3,4 node_modules gitmodules
docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.dev.yml build
docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.dev.yml down
make[2]: Entering directory '/home/giorgio/code/article-store'
make[2]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
make[2]: *** No rule to make target 'w'. Stop.
make[2]: Leaving directory '/home/giorgio/code/article-store'
Makefile:19: recipe for target 'install' failed
...
make[1]: Leaving directory '/home/giorgio/code/article-store'
Makefile:81: recipe for target 'dev' failed
make: *** [dev] Error 2
```
This is because when `install` is called recursively, `MAKEFLAGS` can be seen with an `env` command to evaluate to:
```
MAKEFLAGS=w -j --jobserver-fds=3,4
```
It is set as an environment variable, therefore does not need to be passed on the command line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ help: ## Display this help text
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 install: ## Install dependencies locally
-	make --jobs=2 $(MAKEFLAGS) node_modules gitmodules
+	$(MAKE) node_modules gitmodules
 
 node_modules: package.json package-lock.json
 	npm install
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ help: ## Display this help text
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 install: ## Install dependencies locally
-	make node_modules gitmodules
+	$(MAKE) node_modules gitmodules
 
 node_modules: package.json package-lock.json
 	npm install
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -16,7 +16,7 @@ help: ## Display this help text
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 install: ## Install dependencies locally
-	make --jobs=2 $(MAKEFLAGS) node_modules gitmodules
+	$(MAKE) node_modules gitmodules
```
giorgiosironi, line 19: `$(MAKE)` necessary to pass the `MAKEFLAGS`, otherwise it evaluates to just `MAKEFLAGS=w` rather than `MAKEFLAGS=w -j --jobserver-fds=3,4`.

giorgiosironi: Since the jobserver is the same and has a total, the number of jobs is unnecessary. The subtargets are executed in parallel (between those two, but also with everything else):
```
make[2]: Entering directory '/home/giorgio/code/article-store'
npm install
...
git submodule update --init --recursive
...
touch node_modules
```

thewilkybarkid: When coming from `make dev` rather than `make install`?

thewilkybarkid: Rather speed up `dev` than `install`, so let's go with this. Though these should now be prerequisites?

giorgiosironi: Yes, `make install` doesn't break because there are no preset `MAKEFLAGS` coming from a parent `make --jobs` when invoked (I think).

Could be prerequisites, but making a minimal change to fix `master` first.

thewilkybarkid (approved):

PR #88, opened 2019-12-06
Check for schema:Article and schema:name in post request

nlisgo:
We need to check for that the `@type` includes `http://schema.org/Article` and that the `http://schema.org/name` property exists and has at least one value.

In order to be able to perform these checks we must inspect the expanded JSON-LD object. This ensures that any `@context` is applied to the values in the JSON-LD object.


nlisgo:
Adding tests.

test/routes/add-article.test.ts
```diff
@@ -28,4 +28,18 @@ describe('add article', (): void => {
   it('should throw an error if id is already set', async (): Promise<void> => {
     await expect(makeRequest(createArticle('_:1'))).rejects.toThrow(new createHttpError.Forbidden('Article IDs must not be set (\'_:1\' was given)'));
   });
+
+  it('should throw an error if it is not a schema:Article', async (): Promise<void> => {
+    await expect(makeRequest({
+      '@id': '_:1',
+      '@type': 'http://schema.org/NewsArticle',
```
giorgiosironi, line 37: interesting case, because this is a subtype of `http://schema.org/Article`. Perhaps use something clearly different like https://schema.org/Book?

src/routes/add-article.ts
```diff
@@ -1,16 +1,24 @@
...
+    const [article] = await jsonld.expand(request.body) as JsonLdArray;
+    if (!(article['@type'].includes(schema.Article))) {
+      throw new createHttpError.Forbidden(`Article type must be ${schema.Article} ('${article['@type'].join(', ')}' was given)`);
+    }
     if ('@id' in article) {
       throw new createHttpError.Forbidden(`Article IDs must not be set ('${article['@id']}' was given)`);
     }
+    if (!(schema.name in article) || article[schema.name].length === 0) {
+      throw new createHttpError.Forbidden(`Article must have at least one ${schema.name}`);
+    }
```
giorgiosironi, line 19: the checks seem functionally correct and consistent with existing ones like the one on `@id`

test/routes/add-article.test.ts
thewilkybarkid: Useful to include both cases (as well as one that is both an `Article` _and_ `Book`). RDF inheritance is not something I'm that clear on what we should be doing. Seems like inheritance is not normally understood, but you can choose to understand it. (Declaring it as both an `Article` and `NewsArticle` is the more explicit way).

I'm even less clear on _property_ inheritance...

/cc @stephenwf @tpluscode in case they have any words of wisdom

test/routes/add-article.test.ts
```diff
@@ -28,4 +28,18 @@ describe('add article', (): void => {
...
+      'http://schema.org/name': 'Article _:1',
+    })).rejects.toThrow(new createHttpError.Forbidden('Article type must be http://schema.org/Article (\'http://schema.org/NewsArticle\' was given)'));
+  });
+
+  it('should throw an error if no scheam:name set', async (): Promise<void> => {
```
giorgiosironi: (typo)

test/routes/add-article.test.ts
```diff
@@ -28,4 +28,18 @@ describe('add article', (): void => {
...
+    await expect(makeRequest({
+      '@type': 'http://schema.org/Article',
+    })).rejects.toThrow(new createHttpError.Forbidden('Article must have at least one http://schema.org/name'));
```
giorgiosironi: there are two scenarios to test that the check is catering for: missing field and a field evaluating to an empty string

test/routes/add-article.test.ts
giorgiosironi: (not sure what should happen on a subtype)

test/routes/add-article.test.ts
nlisgo: Have just added these scenarios

test/routes/add-article.test.ts
nlisgo: Interesting to generate this discussion. It was not intended but is valuable. This scenario was just supposed to add a `@type` that was not expected.

thewilkybarkid: For now, this is correctly a failing case (`http://schema.org/Article` needs to be explicit). Might change in the future.

giorgiosironi: My understanding of subtyping is that it's an open world where you can't possibly know all of the subtypes B, C, ... of type A; so to allow subtypes to be passed in you would have to resolve somehow B and understand whether it declares A as its supertype. The client could also indicate `@type` contains both the A and B to make it clear.

giorgiosironi: So checklist is:
- `Book` (refuse as different type)
- both `Article` and `Book` (accept as the first type allows it)
- `NewsArticle` (refuse currently, as it's a subtype)

nlisgo:
Resolving conflicts.

src/routes/add-article.ts
```diff
@@ -1,16 +1,24 @@
...
+import { schema } from 'rdf-namespaces';
 import { AppContext, AppMiddleware } from '../app';
 
 export default (): AppMiddleware => (
   async ({ articles, request, response }: AppContext, next: Next): Promise<void> => {
-    const article = request.body as JsonLdObj;
+    const [article] = await jsonld.expand(request.body) as JsonLdArray;
+    if (!(article['@type'].includes(schema.Article))) {
+      throw new createHttpError.Forbidden(`Article type must be ${schema.Article} ('${article['@type'].join(', ')}' was given)`);
+    }
```
thewilkybarkid: This should probably be on the article repository.

src/routes/add-article.ts
```diff
@@ -1,16 +1,24 @@
...
     if ('@id' in article) {
       throw new createHttpError.Forbidden(`Article IDs must not be set ('${article['@id']}' was given)`);
     }
+    if (!(schema.name in article) || article[schema.name].length === 0) {
+      throw new createHttpError.Forbidden(`Article must have at least one ${schema.name}`);
```
thewilkybarkid: Both the new checks should lead to a `400 Bad Request` error (they're not _valid_, in comparison to providing an `@id` which is not _allowed_).

src/routes/add-article.ts
nlisgo: This check has been moved to `in-memory-articles`. The error thrown there leads to `400 Bad Request`.

src/routes/add-article.ts
nlisgo: This check has been moved to `in-memory-articles`. The error thrown there leads to `400 Bad Request`.

src/errors/article-has-no-title.ts
```diff
@@ -0,0 +1,7 @@
+import { schema } from 'rdf-namespaces';
```
thewilkybarkid: 👍 (should do it elsewhere too)

src/errors/article-has-incorrect-type.ts
```diff
@@ -0,0 +1,5 @@
+export default class ArticleHasIncorrectType extends Error {
+  constructor(type?: string) {
```
thewilkybarkid: Could be more than 1

src/errors/article-has-incorrect-type.ts
```diff
@@ -0,0 +1,5 @@
+export default class ArticleHasIncorrectType extends Error {
```
thewilkybarkid: Maybe `NotAnArticle`?

src/adaptors/in-memory-articles.ts
```diff
@@ -1,15 +1,24 @@
...
   async add(article: JsonLdObj): Promise<void> {
+    if (!('@type' in article) || !(article['@type'].includes(schema.Article))) {
+      throw new ArticleHasIncorrectType(('@type' in article) ? article['@type'] : undefined);
+    }
     if (!('@id' in article)) {
       throw new ArticleHasNoId();
     }
+    if (!(schema.name in article) || article[schema.name].length === 0) {
+      throw new ArticleHasNoTitle();
+    }
```
thewilkybarkid: While this shouldn't expand to other validation steps, suggest we keep this one at the HTTP layer (the other two are required for persistence, this one isn't).

src/errors/article-has-no-title.ts
nlisgo: I have deliberately avoided using it in tests. Thoughts?

src/errors/not-an-article.ts
```diff
@@ -0,0 +1,6 @@
+export default class NotAnArticle extends Error {
+  constructor(type?: string | Array<string>) {
```
thewilkybarkid: I’d avoid doing ‘thing or list of thing’ as it’s awkward to handle (see next line). Better to always have an array argument, or use a varadic (`...`).

src/adaptors/in-memory-articles.ts
```diff
@@ -10,7 +10,7 @@ export default class InMemoryArticles implements Articles {
 
   async add(article: JsonLdObj): Promise<void> {
     if (!('@type' in article) || !(article['@type'].includes(schema.Article))) {
-      throw new NotAnArticle(typeof article['@type'] === 'string' ? [article['@type']] : article['@type']);
+      throw new NotAnArticle([].concat(article['@type'] || []));
```
thewilkybarkid: Looked at using https://www.npmjs.com/package/lodash.castarray too, but also results in `[undefined]` (without `|| []`), so not worth a new library.

test/adaptors/in-memory-articles.test.ts
```diff
@@ -24,10 +25,35 @@ describe('in-memory articles', (): void => {
...
+    });
+
+    expect(await articles.contains('_:1')).toBe(true);
+  });
+
+  it('throws an error if the article does not have correct type', async (): Promise<void> => {
+    const articles = new InMemoryArticles();
+
+    await expect(articles.add({})).rejects.toThrow(new NotAnArticle());
+    await expect(articles.add({ '@type': 'http://schema.org/NewsArticle' })).rejects.toThrow(new NotAnArticle(['http://schema.org/NewsArticle']));
```
thewilkybarkid: These are two separate cases, so we should make them separate tests (otherwise if the first fails we don't know if the second does or not).

test/errors/not-an-article.test.ts
```diff
@@ -0,0 +1,21 @@
...
+  it('should have a message', async (): Promise<void> => {
+    const error = new NotAnArticle();
+
+    expect(error.message).toBe('Article type must be http://schema.org/Article (none was given)');
+  });
+
+  it('may have types', async (): Promise<void> => {
+    const error = new NotAnArticle(['http://schema.org/NewsArticle', 'http://schema.org/Book']);
+
+    expect(error.message).toBe('Article type must be http://schema.org/Article (\'http://schema.org/NewsArticle, http://schema.org/Book\' was given)');
```
thewilkybarkid, line 20: We should make the types accessible and test here.

test/routes/add-article.test.ts
```diff
@@ -23,10 +23,37 @@ describe('add article', (): void => {
...
 
   it('should throw an error if id is already set', async (): Promise<void> => {
     await expect(makeRequest(createArticle('_:1'))).rejects.toThrow(new createHttpError.Forbidden('Article IDs must not be set (\'_:1\' was given)'));
   });
+
+  it('should throw an error if it is not a schema:Article', async (): Promise<void> => {
+    await expect(makeRequest({
+      '@type': 'http://schema.org/NewsArticle',
+      'http://schema.org/name': 'Article _:1',
+    })).rejects.toThrow(new createHttpError.BadRequest('Article type must be http://schema.org/Article (\'http://schema.org/NewsArticle\' was given)'));
```
thewilkybarkid: Was wondering why these were passing. Looks like `toThrow` only compares the message of the error, not the error itself...

thewilkybarkid: `toThrow(createHttpError.BadRequest)` seems to compare the type.

davidcmoulton: Have you tried passing a callback rather than the error object? 
```
toThrow(() => new createHttpError.BadRequest('Article type must be http://schema.org/Article (\'http://schema.org/NewsArticle\' was given)'))
```

test/errors/not-an-article.test.ts
nlisgo: > We should make the types accessible and test here.

Can you give more detail here? Not sure what you mean, sorry.

thewilkybarkid: Meant add a `types` property on the error (like `id` on `ArticleNotFound`).

test/routes/add-article.test.ts
thewilkybarkid: Doesn't take a function.

Wonder if https://www.npmjs.com/package/jest-extended#tothrowwithmessagetype-message should be used everywhere...

giorgiosironi: Wonder if you can compare exceptions at all apart from separately looking at `message` and `name` (which is the type); two otherwise identical exceptions created in two different lines have a different `stack` property:
```
> [(new Error("42")).stack, (new Error("42")).stack]
[
  'Error: 42\n' +
    '    at repl:1:3\n' +
    '    at Script.runInThisContext (vm.js:116:20)\n' +
    '    at REPLServer.defaultEval (repl.js:404:29)\n' +
    '    at bound (domain.js:420:14)\n' +
    '    at REPLServer.runBound [as eval] (domain.js:433:12)\n' +
    '    at REPLServer.onLine (repl.js:715:10)\n' +
    '    at REPLServer.emit (events.js:215:7)\n' +
    '    at REPLServer.EventEmitter.emit (domain.js:476:20)\n' +
    '    at REPLServer.Interface._onLine (readline.js:316:10)\n' +
    '    at REPLServer.Interface._line (readline.js:693:8)',
  'Error: 42\n' +
    '    at repl:1:28\n' +
    '    at Script.runInThisContext (vm.js:116:20)\n' +
    '    at REPLServer.defaultEval (repl.js:404:29)\n' +
    '    at bound (domain.js:420:14)\n' +
    '    at REPLServer.runBound [as eval] (domain.js:433:12)\n' +
    '    at REPLServer.onLine (repl.js:715:10)\n' +
    '    at REPLServer.emit (events.js:215:7)\n' +
    '    at REPLServer.EventEmitter.emit (domain.js:476:20)\n' +
    '    at REPLServer.Interface._onLine (readline.js:316:10)\n' +
    '    at REPLServer.Interface._line (readline.js:693:8)'
]
```

thewilkybarkid: > Wonder if you can compare exceptions at all apart

davidcmoulton: > Doesn't take a function.
> 
Ah, yeah I was thinking of chai. The API is so similar, the differences can be sneaky.

src/adaptors/in-memory-articles.ts
thewilkybarkid: In fact https://github.com/libero/article-store/blob/0ee7221bc9d90d4150abc67f87d733daa02edb2f/src/adaptors/in-memory-articles.ts#L12 requires it to be an array anyway!

thewilkybarkid: Or not quite, odd behaviour if it's a string. Refactoring.

thewilkybarkid: https://github.com/libero/article-store/pull/88/commits/8a743947821b7769303d06f978c2437173662a05

test/routes/add-article.test.ts
nlisgo: Trying to setup `jest-extended` so I can use the `toThrowWithMessage` assertion. Having trouble with setup: https://github.com/libero/article-store/pull/101

test/routes/add-article.test.ts
```diff
@@ -23,10 +23,29 @@ describe('add article', (): void => {
...
+      ...createArticle(),
+      '@type': 'http://schema.org/NewsArticle',
+    })).rejects.toThrow(new createHttpError.BadRequest('Article type must be http://schema.org/Article (\'http://schema.org/NewsArticle\' was given)'));
+  });
+
+  it.each([
+    [undefined],
+    [[]],
+    [{ '@value': null }],
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
thewilkybarkid: Not ideal. No type, or `unknown` doesn't work either.

test/routes/add-article.test.ts
Cooryd: I usually use a `toEqual` assertion when using `.rejects`. Or maybe this could this work? 
```
expect(async () => await makeRequest({
      '@type': 'http://schema.org/NewsArticle',
      'http://schema.org/name': 'Article _:1',
    })).toThrow(new createHttpError.BadRequest('Article type must be http://schema.org/Article (\'http://schema.org/NewsArticle\' was given)'));
```

test/routes/add-article.test.ts
```diff
@@ -34,7 +34,12 @@ describe('add article', (): void => {
     await expect(makeRequest({
       ...createArticle(),
       '@type': 'http://schema.org/NewsArticle',
-    })).rejects.toThrow(new createHttpError.BadRequest('Article type must be http://schema.org/Article (\'http://schema.org/NewsArticle\' was given)'));
+    })).rejects.toBeInstanceOf(createHttpError.BadRequest);
+
+    await expect(makeRequest({
+      ...createArticle(),
+      '@type': 'http://schema.org/NewsArticle',
```
thewilkybarkid: Might be better to extract this to an `article` constant.

src/middleware/error-handler.ts
```diff
@@ -16,9 +18,12 @@ const handleHttpError = (error: HttpError, { response }: Context): void => {
   response.type = 'jsonld';
 };
 
-const toHttpError = (error: UnknownError): HttpError => (
-  error instanceof HttpError ? error : createError(error)
-);
+const toHttpError = (error: UnknownError): HttpError => {
+  if (error instanceof NotAnArticle) {
+    error = new createHttpError.BadRequest(error.message);
+  }
```
thewilkybarkid: This should happen in the `add-article` handler.

test/routes/add-article.test.ts
thewilkybarkid: Turns out it was the use of arrays that stopped `unknown` working (https://github.com/libero/article-store/pull/88/commits/9128fa48158c8efa0a39ab1be915a5538dfdcf09).

thewilkybarkid (approved):

PR #89, opened 2019-12-06
Only use id in create-article name if defined

nlisgo:
Currently the `schema:name` is set to `Article undefined` if `createArticle()` is called.

### DIFF
--- a/test/create-article.ts
+++ b/test/create-article.ts
@@ -3,7 +3,7 @@ import { Iri, JsonLdObj } from 'jsonld/jsonld-spec';
 
 const isNotUndefined = (arg: unknown): boolean => arg !== undefined;
 
-export default (id?: Iri, name = `Article ${id}`): JsonLdObj => (
+export default (id?: Iri, name = id ? `Article ${id}` : 'Article'): JsonLdObj => (
   deepFilter({
     '@id': id,
     '@type': 'http://schema.org/Article',
--- a/test/create-article.ts
+++ b/test/create-article.ts
@@ -3,7 +3,7 @@ import { Iri, JsonLdObj } from 'jsonld/jsonld-spec';
 
 const isNotUndefined = (arg: unknown): boolean => arg !== undefined;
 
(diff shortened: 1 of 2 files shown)


test/create-article.ts
```diff
@@ -3,7 +3,7 @@ import { Iri, JsonLdObj } from 'jsonld/jsonld-spec';
 
 const isNotUndefined = (arg: unknown): boolean => arg !== undefined;
 
-export default (id?: Iri, name = `Article ${id}`): JsonLdObj => (
+export default (id?: Iri, name = id ? `Article ${id}` : 'Article'): JsonLdObj => (
```
giorgiosironi, line 6: would expect a corresponding test that fails before the change and passes after the change?

giorgiosironi: (assuming this is the correct behavior)

nlisgo: I had considered adding a `create-article.test.ts`.

nlisgo: I have added a failing test now and followed up with the fix.

nlisgo:
https://github.com/libero/article-store/pull/89/commits/b1f8fe7b8637133aed23de97d11cb50c08ea8859 should have failed.

https://github.com/libero/article-store/runs/336662411

giorgiosironi:
Weird, does it reproduce locally?

giorgiosironi:
b1f8fe7b8637133aed23de97d11cb50c08ea8859 does fail for me locally.

giorgiosironi:
The [Travis CI build](https://travis-ci.com/libero/article-store/builds/139694115) points to a [merge commit](https://github.com/libero/article-store/commit/0d0b907685f8f95ab6b67b5489c7b3cb40c271ed).

giorgiosironi:
Even though I can diff that commit with `master`, can't download it or open a PR with it. I think it's gone for good.

giorgiosironi (approved):

PR #92, opened 2019-12-06
Remove Travis CI build

giorgiosironi:

Follow up to https://github.com/libero/article-store/pull/74/


giorgiosironi:
Removing the `Travis CI - Pull Request` required check from https://github.com/libero/article-store/settings/branch_protection_rules/11906254 /cc @libero/node-js-reviewers

giorgiosironi:
See no equivalent to a global commit status to check for the `CI` workflow, besides a spurious `.github/workflows/ci.yml` which I don't think is still being produced.

giorgiosironi:
> See no equivalent to a global commit status to check for the CI workflow

AFAIK this cannot be created manually, because the [`GITHUB_TOKEN` provided to fork builds is read only](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token#permissions-for-the-github_token).

giorgiosironi:
Can check all of them if we want, but the list needs to be maintained in case they are renamed, added or removed.

thewilkybarkid (approved):

PR #101, opened 2019-12-10
Avoid using toThrow() and favour separate checks for instanceOf and message property or error

nlisgo:
When checking that an appropriate error is throw we currently need to make 2 passes to check the error type and the error message. The following only checks that the error message is as expected:

```
await expect(makeRequest(createArticle('_:1'))).rejects.toThrow(new createHttpError.ImATeapot('Article IDs must not be set (\'_:1\' was given)'));
```

But the type of message `createHttpError.ImATeapot` is completely ignored.

We would currently need to do something like:

```
await expect(makeRequest(createArticle('_:1'))).rejects.toThrow(createHttpError.Forbidden);
await expect(makeRequest(createArticle('_:1'))).rejects.toThrow('Article IDs must not be set (\'_:1\' was given)');
```
~~But this can be improved by using the `toThrowWithMessage` check available with the `jest-extended` library. See: https://www.npmjs.com/package/jest-extended#tothrowwithmessagetype-message~~

`toThrowWithMessage` does not currently work with promises: https://github.com/jest-community/jest-extended/issues/219


jest.config.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
+  setupFilesAfterEnv: [
+    'jest-extended',
+  ],
```
thewilkybarkid: Try enabling in the test file itself (like https://github.com/thewilkybarkid/article-store/blob/d0a1f2d1adfe67f5f665af36e5f3603e54351a1f/test/middleware/error-handler.test.ts#L2).

nlisgo: Did the trick!

nlisgo:
We now have this issue: https://github.com/jest-community/jest-extended/issues/219

test/adaptors/in-memory-articles.test.ts
```diff
@@ -27,7 +27,8 @@ describe('in-memory articles', (): void => {
   it('throws an error if the article does not have an ID', async (): Promise<void> => {
     const articles = new InMemoryArticles();
 
-    await expect(articles.add({})).rejects.toThrow(new ArticleHasNoId());
+    await expect(articles.add({})).rejects.toBeInstanceOf(ArticleHasNoId);
+    await expect(articles.add({})).rejects.toHaveProperty('message', 'Article must have an ID');
```
thewilkybarkid: Probably better not to check the message here in case it's changed.

test/adaptors/in-memory-articles.test.ts
```diff
@@ -41,7 +42,8 @@ describe('in-memory articles', (): void => {
   it('throws an error if the article is not found', async (): Promise<void> => {
     const articles = new InMemoryArticles();
 
-    await expect(articles.get('_:1')).rejects.toThrow(new ArticleNotFound('_:1'));
+    await expect(articles.get('_:1')).rejects.toBeInstanceOf(ArticleNotFound);
+    await expect(articles.get('_:1')).rejects.toHaveProperty('message', 'Article _:1 could not be found');
```
thewilkybarkid: Like with https://github.com/libero/article-store/pull/101#discussion_r356060843, we probably don't want to check the message. But, we should be able to check `toHaveProperty('id', ' _:1')`.

thewilkybarkid (approved):

PR #102, opened 2019-12-10
Use Hydra Analyser Docker container

thewilkybarkid:
Removes quite a few NPM dependencies.

### DIFF
--- a/.github/analyse.sh
+++ b/.github/analyse.sh
@@ -9,4 +9,4 @@ trap finish EXIT
 
 make start wait-healthy
 
-make exec tty= command="npx hydra-validator analyse http://localhost:8080/"
+docker run --rm --network host hydrofoil/hydra-analyser:0.2.0 http://localhost:8080/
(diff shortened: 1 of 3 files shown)


.github/analyse.sh
```diff
@@ -9,4 +9,4 @@ trap finish EXIT
 
 make start wait-healthy
 
-make exec tty= command="npx hydra-validator analyse http://localhost:8080/"
+docker run --rm --network host hydrofoil/hydra-analyser:0.2.0 http://localhost:8080/
```
thewilkybarkid, line 12: Presume Dependabot will never be able to find this. Related to https://github.com/libero/publisher/issues/350#issuecomment-562026405, might be able to move it to Docker-Compose, and when support is available for that it can be updated.

giorgiosironi: Should it be `docker-compose.dev.yml` rather than `docker-compose.yml` in that case?

thewilkybarkid: Maybe, or a CI one. Not going to do it here anyway.

giorgiosironi (approved):
Works locally as well

PR #105, opened 2019-12-10
Disable @typescript-eslint/no-explicit-an rule for tests

nlisgo:
There are instances where in test scenarios we want to be able to ensure that we can handle different cases that mean we are passing `undefined` or other types to the same parameter.

See: https://github.com/libero/article-store/pull/88/files#r355966492

Rather than setting a `eslint-disable-next-line` note in each instance we should just disable the rule for all test files.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,16 @@ module.exports = {
     'plugin:import/typescript',
     'plugin:jest/recommended',
   ],
+  overrides: [
+    {
+      files: [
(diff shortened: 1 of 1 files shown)


thewilkybarkid:
Turns out the use case for this (Jest's `test.each`) has a proper solution (https://github.com/libero/article-store/pull/88/commits/9128fa48158c8efa0a39ab1be915a5538dfdcf09). There might be other cases where we can't avoid this, but for now...

PR #104, opened 2019-12-10
Add Hypertest scenarios

thewilkybarkid:
Adds functional tests of the Hydra API.

PR #111, opened 2019-12-12
Allow empty response bodies

thewilkybarkid:
Koa [sets the body of a response as the status message](https://github.com/koajs/koa/blob/ed84ee50da8ae3cd08056f944d061e00d06ed87f/lib/application.js#L239-L251) if it hasn't been explicitly set and it's not a regular empty response (eg `204 No Content`). This prevents it from happening.

(This is expected behaviour of Koa, but not of this app, so I'm calling it a bug.)

Extracted from https://github.com/libero/article-store/pull/104.

### DIFF
--- a/src/middleware/empty-response.ts
+++ b/src/middleware/empty-response.ts
@@ -1,16 +1,20 @@
-import { Context, Middleware, Next } from 'koa';
+import {
+  Context, Middleware, Next, Response,
+} from 'koa';
+
+const makeResponseEmpty = (response: Response): void => {
+  response.body = '';
+  response.remove('Content-Length');
+  response.remove('Content-Type');
+  response.remove('Transfer-Encoding');
+};
 
 export default (): Middleware => (
   async ({ response }: Context, next: Next): Promise<void> => {
     await next();
 
-    if (response.body !== undefined) {
-      return;
+    if (response.body === undefined) {
+      makeResponseEmpty(response);
     }
-
-    response.body = '';
-    response.remove('Content-Length');
-    response.remove('Content-Type');
-    response.remove('Transfer-Encoding');
   }
 );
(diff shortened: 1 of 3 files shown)


src/middleware/empty-response.ts
```diff
@@ -0,0 +1,16 @@
...
+    await next();
+
+    if (response.body !== undefined) {
+      return;
+    }
+
+    response.body = '';
+    response.remove('Content-Length');
+    response.remove('Content-Type');
+    response.remove('Transfer-Encoding');
```
thewilkybarkid: `Content-Language`, `Content-Encoding`, `Content-Disposition` and `Content-Location` too?

thewilkybarkid: Or maybe this middleware shouldn't care beyond `Content-Length` and `Content-Type`?

test/middleware/empty-response.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  it('removes redundant headers if no body has been set', async (): Promise<void> => {
+    const next = async ({ response }: Context): Promise<void> => {
+      response.set('Content-Length', '100');
+      response.set('Content-Type', 'content/type');
+      response.set('Transfer-Encoding', 'identity');
+    };
+
+    const response = await makeRequest(next);
+
+    expect(response.get('Content-Length')).toHaveLength(0);
```
giorgiosironi: any perceivable difference between the header being missing and having length 0?

thewilkybarkid: `Koa.Response` returns an empty string, would like to find a better way to check the header doesn't exist at all.

src/middleware/empty-response.ts
```diff
@@ -0,0 +1,16 @@
+import { Context, Middleware, Next } from 'koa';
+
+export default (): Middleware => (
+  async ({ response }: Context, next: Next): Promise<void> => {
+    await next();
+
+    if (response.body !== undefined) {
```
giorgiosironi: isn't it clearer with a positive check? I get early returns, but the special case (undefined body) is here put into the main part of the function rather than the body of this `if`

giorgiosironi (approved):

PR #113, opened 2019-12-12
Postgres

nlisgo:
In order to support the requirement of persistence of articles we need to setup `postgres` and handle the setup of the database table.

This POC PR introduces `@libero/migrator` and has an initial `migration` script which handles the setup of the database table by exposing a `cli` command.

With the `Knex` library configured we can accept post requests that persist entries to the database and at the moment retrieve the count from the database table on the `article-list` query. We to introduce an async iterator to handle the retrieval al the list of articles which can be done on `in-memory-articles` initially.


nlisgo:
Closing in favour of https://github.com/libero/article-store/pull/116

PR #114, opened 2019-12-12
Article list async iterable

nlisgo:
This implements an async iterator on implementations of the `Articles` class.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,21 @@ module.exports = {
   root: true,
   rules: {
     'no-undef': 'off',
+    'no-restricted-syntax': [
+      'error',
+      {
+        selector: 'ForInStatement',
+        message: 'for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array.',
+      },
+      {
+        selector: 'LabeledStatement',
+        message: 'Labels are a form of GOTO; using them makes code confusing and hard to maintain and understand.',
+      },
+      {
+        selector: 'WithStatement',
+        message: '`with` is disallowed in strict mode because it makes code impossible to predict and optimize.',
+      },
+    ],
   },
   settings: {
     'import/resolver': {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,21 +20,6 @@ module.exports = {
   root: true,
   rules: {
     'no-undef': 'off',
-    'no-restricted-syntax': [
-      'error',
-      {
-        selector: 'ForInStatement',
-        message: 'for..in loops iterate over the entire prototype chain, which is virtually never what you want. Use Object.{keys,values,entries}, and iterate over the resulting array.',
-      },
-      {
-        selector: 'LabeledStatement',
-        message: 'Labels are a form of GOTO; using them makes code confusing and hard to maintain and understand.',
-      },
-      {
-        selector: 'WithStatement',
-        message: '`with` is disallowed in strict mode because it makes code impossible to predict and optimize.',
-      },
-    ],
   },
   settings: {
     'import/resolver': {
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.5",
+    "async-iterator-all": "^1.0.0",
     "content-type": "^1.0.4",
     "format-link-header": "^2.1.0",
     "http-errors": "^1.7.3",
--- a/package.json
+++ b/package.json
@@ -3,10 +3,10 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.5",
-    "async-iterator-all": "^1.0.0",
     "content-type": "^1.0.4",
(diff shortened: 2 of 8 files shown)


nlisgo:
There is currently a rule in place which discourages the use of `for await` which has resulted in the lint error:

```
/app/src/routes/article-list.ts
  11:5  error  iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations  no-restricted-syntax
```

stephenwf:
@nlisgo Might be worth checking if the error is right when it says that babel will add the regenerator runtime. I think if the output JS files contain the regenerator "polyfill" for async/await it should be removed. Node 8+ supports async/await natively, and it is much quicker!

thewilkybarkid:
I did do a bit of reading recently and there are complaints that AirBnB's style has this as it introduces bad practice, but the problem is that it's someone else's coding standard that isn't necessarily going to meet others' requirements. Need to look at opting in to this rules soon rather than just adopting something existing and having to disable things.

nlisgo:
> @nlisgo Might be worth checking if the error is right when it says that babel will add the regenerator runtime. I think if the output JS files contain the regenerator "polyfill" for async/await it should be removed. Node 8+ supports async/await natively, and it is much quicker!

Natively supported.

.eslintrc.js
```diff
@@ -20,6 +20,21 @@ module.exports = {
...
+      },
+      {
+        selector: 'LabeledStatement',
+        message: 'Labels are a form of GOTO; using them makes code confusing and hard to maintain and understand.',
+      },
+      {
+        selector: 'WithStatement',
+        message: '`with` is disallowed in strict mode because it makes code impossible to predict and optimize.',
+      },
+    ],
```
nlisgo: We should consider splitting this out in to another PR. Perhaps taking on the challenge of maintaining our own standards rather than inheriting someone else's. (see: https://github.com/libero/article-store/pull/114#issuecomment-565065488)

nlisgo: I'm not sure if we can simply disable that one rule, I think we are going to need to explicitly list all of the rules we want to follow in `no-restricted-syntax`.

thewilkybarkid: To delay the linting change, let’s try something like https://www.npmjs.com/package/async-iterator-all

src/routes/article-list.ts
```diff
@@ -20,7 +21,7 @@ export default (): AppMiddleware => (
       },
       [hydra.totalItems]: await articles.count(),
       [hydra.member]: {
-        '@list': [...articles],
+        '@list': [...await all(articles)],
```
thewilkybarkid: Micro-optimization (and doesn't make a difference given the current implementation is synchronous), but we should group promises into a single `await` when they're not ordered.

test/routes/add-article.test.ts
```diff
@@ -21,9 +22,14 @@ describe('add article', (): void => {
     const articles = new InMemoryArticles();
     const response = await makeRequest(createArticle(), undefined, articles);
 
+    const list = [];
+    for await (const a of articles) {
+      list.push(a);
+    }
```
thewilkybarkid: Redundant now.

package.json
```diff
@@ -3,6 +3,7 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.5",
+    "async-iterator-all": "^1.0.0",
```
thewilkybarkid: Useful to add types for this. Fancy having a crack at it?

In https://github.com/libero/article-store/blob/60161cffe176be651ef74af5af02c56d0c3a19ab/src/routes/article-list.ts#L11 `count` is correctly a number, but `list` is `any`.

thewilkybarkid: (Though maybe there is already a typed alternative.)

thewilkybarkid: Can't see one; unclear, but looking like the library has been replaced by https://www.npmjs.com/package/it-all (same author).

nlisgo: > Fancy having a crack at it?

I will have a go at creating a PR for https://github.com/DefinitelyTyped/DefinitelyTyped

test/routes/add-article.test.ts
```diff
@@ -20,11 +21,12 @@ describe('add article', (): void => {
...
     const articles = new InMemoryArticles();
     const response = await makeRequest(createArticle(), undefined, articles);
+    const [list, count] = await Promise.all([all(articles), articles.count()]);
 
     expect(response.status).toBe(201);
     expect(response.get('Location')).toBe('http://example.com/path-to/article-list');
-    expect(await articles.count()).toBe(1);
-    expect([...articles][0]['http://schema.org/name']).toEqual([{ '@value': 'Article' }]);
+    expect(count).toBe(1);
+    expect(list[0]['http://schema.org/name']).toEqual([{ '@value': 'Article' }]);
```
thewilkybarkid: I'd actually keep the individual `await`s here to isolate what each one does (eg if the iterator fails for some reason, can be good to know the previous `expect`s were fine).

package.json
nlisgo: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41119

thewilkybarkid (approved):

PR #115, opened 2019-12-13
Use RDF/JS DatasetCore

thewilkybarkid:
Uses RDF/JS [`DatasetCore`](https://rdf.js.org/dataset-spec/#datasetcore-interface) instead of trying to use plain JSON-LD.

A `dataset` property is added to requests and responses; the former's populated when a request contains JSON-LD, then the latter is populated by the handlers and used to populate the response body.

An article represented in a dataset; the ID of the dataset itself is the ID of the 'root' resource, which has to be a `schema:Article`.

Follow up to https://github.com/libero/article-store/pull/64.


src/adaptors/in-memory-articles.ts
```diff
@@ -1,48 +1,44 @@
-import { Iri, JsonLdObj } from 'jsonld/jsonld-spec';
-import { schema } from 'rdf-namespaces';
+import {
+  BlankNode, DatasetCore, Quad, Quad_Object as QuadObject,
```
thewilkybarkid, line 2: Rename is a little annoying, eslint complains even though it's defined outside...

src/middleware/api-documentation-link.ts
```diff
@@ -1,14 +1,14 @@
...
 import url from 'url';
+import { hydra } from '../namespaces';
 
 export default (path: string): Middleware => (
   async ({ request, response }: Context, next: Next): Promise<void> => {
     await next();
 
     const link = {
-      rel: hydra.apiDocumentation,
+      rel: hydra.apiDocumentation.value,
```
thewilkybarkid: Not sure should be doing this anyway, it's the same as the RDF term, rather than being RDF itself.

src/middleware/dataset.ts
```diff
@@ -0,0 +1,14 @@
+import { Next } from 'koa';
+// eslint-disable-next-line import/no-cycle
+import { AppContext, AppMiddleware } from '../app';
```
thewilkybarkid: https://github.com/benmosher/eslint-plugin-import/issues/1453. But I have an idea on a refactor to get around it.

src/middleware/jsonld.ts
```diff
@@ -1,17 +1,31 @@
+import ParserJsonld from '@rdfjs/parser-jsonld';
+import SerializerJsonld from '@rdfjs/serializer-jsonld-ext';
```
thewilkybarkid, line 2: These are both streaming, which is a bit redundant given the dataset isn't. Not found a parser/serializer that isn't streaming yet.

src/namespaces.ts
```diff
@@ -0,0 +1,6 @@
+import namespace from '@rdfjs/namespace';
+
+export const hydra = namespace('http://www.w3.org/ns/hydra/core#');
+export const owl = namespace('http://www.w3.org/2002/07/owl#');
+export const rdf = namespace('http://www.w3.org/1999/02/22-rdf-syntax-ns#');
+export const schema = namespace('http://schema.org/');
```
thewilkybarkid: Think I remember seeing a library that could replace this. Useful to have one that knows the terms to avoid typos.

src/routes/add-article.ts
```diff
@@ -1,27 +1,33 @@
...
+import { rdf, schema } from '../namespaces';
 
 export default (): AppMiddleware => (
   async ({ articles, request, response }: AppContext, next: Next): Promise<void> => {
-    const [article] = await jsonld.expand(request.body) as JsonLdArray;
-    if ('@id' in article) {
-      throw new createHttpError.Forbidden(`Article IDs must not be set ('${article['@id']}' was given)`);
+    const graph = clownface({ dataset: request.dataset });
+
+    const foundArticles = graph.has(rdf.type, schema.Article);
```
thewilkybarkid: Thinking about it, can probably just use `dataset.match()`.

src/routes/add-article.ts
```diff
@@ -1,27 +1,33 @@
...
     }
-    if (!(schema.name in article) || article[schema.name].length === 0) {
-      throw new createHttpError.BadRequest(`Article must have at least one ${schema.name}`);
+
+    if (foundArticles.terms.length === 0) {
+      throw new createHttpError.BadRequest(`No ${schema.Article.value} found`);
     }
 
-    article['@id'] = `_:${uniqueString()}`;
+    const id = foundArticles.terms[0];
```
thewilkybarkid: Need a check to make sure this a blank node, then migrate all the statements to a new unique blank node (will be a named node in the future).

src/routes/api-documentation.ts
```diff
@@ -1,140 +1,141 @@
...
+      quad(hydra.Collection, hydra.supportedProperty, collectionMember),
+      quad(collectionMember, rdf.type, hydra.SupportedProperty),
+      quad(collectionMember, hydra.title, literal('Members of this collection', 'en')),
+      quad(collectionMember, hydra.property, hydra.member),
+      quad(hydra.member, rdf.type, rdf.Property),
+      quad(hydra.member, rdf.type, hydra.Link),
+      quad(collectionMember, hydra.required, toRdf(true)),
+      quad(collectionMember, hydra.readable, toRdf(false)),
+      quad(collectionMember, hydra.writeable, toRdf(false)),
+    );
```
thewilkybarkid: See also https://github.com/libero/article-store/pull/64#discussion_r352024083.

src/app.ts
```diff
@@ -22,22 +31,22 @@ export default (
...
   app.use(emptyResponse());
-  app.use(bodyParser({
-    extendTypes: {
-      json: ['application/ld+json'],
-    },
-  }));
   app.use(cors({
     exposeHeaders: ['Link'],
   }));
+  app.use(dataset());
```
thewilkybarkid: Could do with a better name

src/routes/add-article.ts
```diff
@@ -1,27 +1,30 @@
...
     }
-    if (!(schema.name in article) || article[schema.name].length === 0) {
-      throw new createHttpError.BadRequest(`Article must have at least one ${schema.name}`);
+
+    if (foundArticles.size === 0) {
+      throw new createHttpError.BadRequest(`No ${schema.Article.value} found`);
     }
 
-    article['@id'] = `_:${uniqueString()}`;
+    const id = [...foundArticles][0].subject;
```
thewilkybarkid: This ID needs to be validated that it's a blank node, then replace with a new unique ID.

src/articles.ts
```diff
@@ -1,13 +1,13 @@
-import { Iri, JsonLdObj } from 'jsonld/jsonld-spec';
+import { BlankNode, DatasetCore } from 'rdf-js';
 
-interface Articles extends Iterable<JsonLdObj> {
-  add(article: JsonLdObj): Promise<void>;
+interface Articles extends Iterable<[BlankNode, DatasetCore]> {
+  set(id: BlankNode, article: DatasetCore): Promise<void>;
```
thewilkybarkid, line 4: Renamed `add` to `set` as we have to separately include the ID. Makes it a bit more persistence-orientated too, and might remove the need for a unit of work when introduces a database.

src/routes/add-article.ts
```diff
@@ -1,27 +1,34 @@
...
     }
 
-    article['@id'] = `_:${uniqueString()}`;
+    if (request.dataset.match(id, schema.name).size === 0) {
+      throw new createHttpError.BadRequest(`Article must have at least one ${schema.name.value}`);
+    }
 
     try {
-      await articles.add(article);
+      await articles.set(id, request.dataset);
```
thewilkybarkid: The ID here needs to be replaced with a new unique one (this will be a named node in the future).

src/articles.ts
```diff
@@ -1,13 +1,13 @@
-import { Iri, JsonLdObj } from 'jsonld/jsonld-spec';
+import { BlankNode, DatasetCore } from 'rdf-js';
 
-interface Articles extends Iterable<JsonLdObj> {
-  add(article: JsonLdObj): Promise<void>;
+interface Articles extends Iterable<[BlankNode, DatasetCore]> {
```
thewilkybarkid: `BlankNode` will be replaced with `NamedNode` in the future.

src/routes/article-list.ts
```diff
@@ -1,29 +1,37 @@
...
+      quad(articleList, rdf.type, hydra('Collection')),
+      quad(articleList, hydra.title, literal('List of articles', 'en')),
+      quad(articleList, hydra.manages, manages),
+      quad(manages, hydra.property, rdf.type),
+      quad(manages, hydra.object, schema.Article),
+      quad(articleList, hydra.totalItems, toRdf(await articles.count())),
+    ];
+
+    for (const [id, article] of articles) {
+      quads.push(quad(articleList, hydra.member, id), ...article);
```
thewilkybarkid: Suspect `...article` might have some collisions. Won't be embedding them when they're separate resources anyway.

src/routes/entry-point.ts
```diff
@@ -1,23 +1,28 @@
...
-        '@id': router.url(Routes.ArticleList),
-        '@type': hydra.Collection,
-      },
-    };
-    response.type = 'jsonld';
+  async ({
+    dataFactory: { quad, literal, namedNode }, request, response, router,
+  }: AppContext, next: Next): Promise<void> => {
+    const articleList = namedNode(url.resolve(request.origin, router.url(Routes.ArticleList)));
+    const entryPoint = namedNode(url.resolve(request.origin, router.url(Routes.EntryPoint)));
```
thewilkybarkid: It'd be better for all these named nodes to be relative, then make absolute when writing the response body.

test/middleware.ts
```diff
@@ -1,18 +1,21 @@
 import { Middleware, ParameterizedContext, Response } from 'koa';
+import { DatasetCore } from 'rdf-js';
 import { AppContext, AppState } from '../src/app';
 
 export type Next<StateT = AppState, CustomT = AppContext> = (
   context: ParameterizedContext<StateT, CustomT>
 ) => Promise<void>;
 
+export type DatasetResponse = Response & { dataset: DatasetCore };
```
thewilkybarkid: This is a bit annoying.

src/namespaces.ts
tpluscode: `@zazuko/rdf-vocabularies` does that to some extent

src/middleware/jsonld.ts
tpluscode: The streaming is both about the input, which dataset maybe is not, bu also about streaming the output content ;)

src/routes/api-documentation.ts
tpluscode: I do simply write all my API Documentation in turtle and parse at app start.

I see you already have `clownface` here. This could become more approachable

```js
import cf = require('clownface')
import dataset = require('rdf-dataset-indexed')

const apiDocumentation = cf({
  dataset: dataset(),
  term: namedNode(url.resolve(request.origin, router.url(Routes.ApiDocumentation)))
})

// individual triples are boring
apiDocumentation.addOut(rdf.type, hydra.ApiDocumentation)
apiDocumentation.addOut(hydra.entrypoint, entryPoint)

// but you can create deep graphs with callbacks
apiDocumentation.addOut(hydra.supportedClass, schema.EntryPoint, entrypoint => {
  entrypoint.addOut(rdf.type, hydra.Class)
  entrypoint.addOut(hydra.title, literal('API entry point', 'en'))

  entrypoint.addOut(rdf.type, hydra.supportedOperation, entryPointGet => {
    entryPointGet.addOut(rdf.type, hydra.Operation)
  })
})
```

test/middleware.ts
tpluscode: I think module augmentation might help. Here's how we extend express' Request and Response: https://github.com/zazuko/data-cube-curation/blob/master/api/typings/express/index.d.ts

src/namespaces.ts
thewilkybarkid: [`@ontologies/schema`](https://www.npmjs.com/package/@ontologies/schema) etc was what I was thinking of. Just tried but ran into https://github.com/ontola/ontologies/issues/3.

src/app.ts
```diff
@@ -14,6 +15,14 @@ export type AppState = DefaultState;
 
 export type AppContext = RouterContext<AppState, {
   articles: Articles;
+  dataFactory: DataFactory;
+  datasetFactory: DatasetCoreFactory;
+  request: {
```
giorgiosironi: I guess the tests have some reset mechanism for these `request` and `response`, or they build a new `AppContext` every time?

src/articles.ts
giorgiosironi: Elaborate with an example on the difference?

giorgiosironi: e.g. does the `NamedNode` have a URI?

src/routes/add-article.ts
```diff
@@ -1,27 +1,34 @@
...
+
+    const id = [...foundArticles][0].subject;
+
+    if (id.termType !== 'BlankNode') {
+      throw new createHttpError.BadRequest(`Article must have a blank node identifier (${id.value} given)`);
     }
 
-    article['@id'] = `_:${uniqueString()}`;
+    if (request.dataset.match(id, schema.name).size === 0) {
+      throw new createHttpError.BadRequest(`Article must have at least one ${schema.name.value}`);
```
giorgiosironi: should this validation live in the lower layer?

src/articles.ts
thewilkybarkid: `NamedNode` is an IRI (=unicode URI). Since we're using Hydra and declare it as a Hydra resource, that means it would also be dereferenceable (ie _also_ a URL). `BlankNode` is just a local identifier, in this case inside the article list.

src/app.ts
thewilkybarkid: Context is fresh for every request handled.

src/middleware/jsonld.ts
thewilkybarkid: Yeah, mean in terms out parser output and serialiser input!

src/routes/api-documentation.ts
thewilkybarkid: Interesting, will have a go. 👍

giorgiosironi: The nesting via callbacks is interesting here in terms of readability.

test/adaptors/in-memory-articles.test.ts
```diff
@@ -1,116 +1,126 @@
...
     const articles = new InMemoryArticles();
+    const id = blankNode();
 
-    await expect(articles.add({
-      ...createArticle('_:1'),
-      '@type': 'http://schema.org/NewsArticle',
-    })).rejects.toThrow(new NotAnArticle(['http://schema.org/NewsArticle']));
+    const article = createArticle(id)
+      .delete(quad(id, rdf.type, schema.Article))
+      .add(quad(id, rdf.type, schema.NewsArticle));
```
giorgiosironi: pass `schema.NewsArticle` as an optional argument rather than deleting the other type?

package.json
```diff
@@ -3,41 +3,44 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.5",
+    "@rdfjs/data-model": "^1.1.2",
+    "@rdfjs/dataset": "^1.0.1",
+    "@rdfjs/namespace": "^1.1.0",
+    "@rdfjs/parser-jsonld": "^1.2.0",
+    "@rdfjs/serializer-jsonld-ext": "^2.0.0",
```
thewilkybarkid: Adding types in https://github.com/rdfjs-base/serializer-jsonld-ext/pull/2.

thewilkybarkid, line 9: Adding types in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41095.

package.json
```diff
@@ -3,41 +3,44 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.5",
+    "@rdfjs/data-model": "^1.1.2",
+    "@rdfjs/dataset": "^1.0.1",
+    "@rdfjs/namespace": "^1.1.0",
+    "@rdfjs/parser-jsonld": "^1.2.0",
```
thewilkybarkid, line 8: Adding types in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41093.

package.json
```diff
@@ -3,41 +3,44 @@
   "dependencies": {
     "@koa/cors": "^3.0.0",
     "@koa/router": "^8.0.5",
+    "@rdfjs/data-model": "^1.1.2",
+    "@rdfjs/dataset": "^1.0.1",
+    "@rdfjs/namespace": "^1.1.0",
```
thewilkybarkid, line 7: Adding types in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41082.

package.json
```diff
@@ -3,41 +3,44 @@
...
     "http-errors": "^1.7.3",
-    "jsonld": "^2.0.1",
     "koa": "^2.11.0",
-    "koa-bodyparser": "^4.2.1",
     "koa-compose": "^4.1.0",
     "koa-logger": "^3.2.1",
-    "rdf-namespaces": "^1.6.0",
-    "unique-string": "^2.0.0"
+    "p-event": "^4.1.0",
+    "rdf-dataset-ext": "^1.0.0",
```
thewilkybarkid, line 19: Adding types in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41089.

src/index.ts
```diff
@@ -8,6 +8,7 @@ import Routes from './routes';
 const articles = new InMemoryArticles();
 const router = createRouter();
 const apiDocumentationPath = router.url(Routes.ApiDocumentation);
+const datasetFactory = { dataset: createDataset };
```
thewilkybarkid: Refs https://gitter.im/rdfjs/public?at=5df8f727e47fb31eb7cadfc0.

test/create-article.ts
```diff
@@ -1,10 +1,10 @@
 import { blankNode, literal, quad } from '@rdfjs/data-model';
-import { dataset } from '@rdfjs/dataset';
 import { DatasetCore, Quad_Subject as QuadSubject } from 'rdf-js';
+import datasetFactory from '../src/dataset-factory';
```
thewilkybarkid: Not sure why destructuring no longer worked.

tpluscode: Did the export change to the `export = X` form? It is not possible to destructure commonjs-style imports when importing with ES module

thewilkybarkid: Nope: https://github.com/libero/article-store/blob/2e7704d027f585282e51c66b6f4e7acf3c7803cc/src/dataset-factory.ts#L4-L6

tpluscode: Ah, the same is indeed true for default export. This is not in fact object destructuring, despite similar appearance. Cue [this SO answer](https://stackoverflow.com/a/43987935)

thewilkybarkid: Aha, good find.

Anyhow, ignoring as https://github.com/rdf-ext/rdf-dataset-indexed/pull/3#discussion_r359231278 will remove the need for this layer.

thewilkybarkid:
Waiting on some library changes still.

src/routes/add-article.ts
```diff
@@ -30,8 +30,8 @@ export default (): AppMiddleware => (
       throw new createHttpError.BadRequest(`Article must have a blank node identifier (${id.value} given)`);
     }
 
-    if (request.dataset.match(id, schema.name).size === 0) {
-      throw new createHttpError.BadRequest(`Article must have at least one ${schema.name.value}`);
+    if (request.dataset.match(id, schema('name')).size === 0) {
+      throw new createHttpError.BadRequest(`Article must have at least one ${schema('name').value}`);
```
thewilkybarkid: This had to be reverted due to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41082#discussion_r360391625.

giorgiosironi:
Build seems to be stuck not executing new commits :thinking:

giorgiosironi:
Attempting a push to verify whether we have some problems on Github Actions.

thewilkybarkid:
There's a merge conflict so it can't build.

giorgiosironi:
Ah, wasn't clear to me that it was normally [using the merge commit](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#pull-request-event-pull_request).

thewilkybarkid:
(Travis also behaved like this.)

src/routes/api-documentation.ts
```diff
@@ -1,140 +1,145 @@
...
+
+      entryPoint.addOut(hydra.supportedProperty, (name: Clownface): void => {
+        name.addOut(rdf.type, hydra.SupportedProperty);
+        name.addOut(hydra.title, literal('Name', 'en'));
+        name.addOut(hydra.property, schema('name'), (property: Clownface): void => {
+          property.addOut(rdf.type, rdf.Property);
+        });
+        name.addOut(hydra.required, true);
+        name.addOut(hydra.readable, true);
+        name.addOut(hydra.writeable, toRdf(false));
```
thewilkybarkid, line 48: Won't be needed when https://github.com/rdf-ext/clownface/pull/23 is released.

src/adaptors/in-memory-articles.ts
thewilkybarkid: Tried setting `'@typescript-eslint/camelcase': ['error', { ignoreImports: true }]`, stops the import being flagged as an error but any further usage still is...

src/articles.ts
thewilkybarkid: Extracted to https://github.com/libero/article-store/pull/143.

thewilkybarkid:
Extracted the introduction of types in https://github.com/libero/article-store/pull/146, and made a couple of improvements there too.

src/adaptors/in-memory-articles.ts
giorgiosironi: Wonder why just this name follows a different style (than `BlankNode` for example)?

src/middleware/jsonld.ts
```diff
@@ -1,26 +1,39 @@
...
-  Context, Middleware, Next, Response,
+  DefaultState, Middleware, Next, Response,
 } from 'koa';
+import pEvent from 'p-event';
+import { fromStream, toStream } from 'rdf-dataset-ext';
+import { DatasetContext } from './dataset';
 
-const isJsonLd = (response: Response): boolean => response.is('jsonld') && typeof response.body === 'object';
+const responseHasContent = (response: Response): boolean => (
+  response.body || response.status === constants.HTTP_STATUS_NO_CONTENT
```
giorgiosironi, line 14: there is content either if there is a body or if the status is no content?

src/middleware/jsonld.ts
```diff
@@ -1,26 +1,39 @@
...
+  const contentType = {
+    type: 'application/ld+json',
+    parameters: { profile: 'http://www.w3.org/ns/json-ld#compacted' },
+  };
+  const parser = new ParserJsonld();
+  const serializer = new SerializerJsonld({ compact: true, context });
+
+  return async ({ request, response }: DatasetContext, next: Next): Promise<void> => {
+    if (request.is('jsonld')) {
+      request.dataset = await fromStream(request.dataset, parser.import(request.req));
```
giorgiosironi: weird signature where the dataset has to be passed in and then is reassigned. Probably can't do much about it

src/routes/add-article.ts
```diff
@@ -1,33 +1,51 @@
+import clownface from 'clownface';
```
giorgiosironi, line 1: who names these things :older_man: :cloud:

src/routes/add-article.ts
```diff
@@ -1,33 +1,51 @@
...
-    article['@id'] = termToString(newId);
+
+    [...request.dataset].forEach((originalQuad: Quad): void => {
+      let newQuad: Quad;
+      if (originalQuad.subject.equals(id)) {
+        newQuad = quad(newId, originalQuad.predicate, originalQuad.object, originalQuad.graph);
+      } else if (originalQuad.object.equals(id)) {
+        newQuad = quad(originalQuad.subject, originalQuad.predicate, newId, originalQuad.graph);
+      } else {
+        return;
```
giorgiosironi, line 41: ignores silently all data in the graph that is not associated to the article id in the role of subject or object?

src/adaptors/in-memory-articles.ts
thewilkybarkid: https://gitter.im/rdfjs/public?at=5e0e07da04a1900497f07c9e (not sure why)

src/middleware/jsonld.ts
thewilkybarkid: Probably don't need the reassignment...

src/routes/add-article.ts
thewilkybarkid: Leaves them as is. Other option is to create a new dataset, so there's always an add.

src/routes/article-list.ts
```diff
@@ -1,34 +1,47 @@
...
+    const listPromise = all(articles)
+      .then((list): void => {
+        list.forEach(([id, article]: [BlankNode, DatasetCore]): void => {
+          graph.addOut(hydra.member, id);
+          addAll(graph.dataset, article);
+        });
+      });
+
+    const countPromise = articles.count()
+      .then((count): void => {
```
giorgiosironi, line 31: use of multiple Promises for parallelism rather than `await`ing for their result?

thewilkybarkid: Yeah, micro-optimisation but means processing based on those values can happen immediately rather than waiting for both calls to finish.

test/context.ts
```diff
@@ -1,21 +1,26 @@
...
 import { AppContext } from '../src/app';
 import Articles from '../src/articles';
+import dataFactory, { dataset } from '../src/data-factory';
+import { WithDataset } from '../src/middleware/dataset';
 
 export type ErrorListener = (error: UnknownError, context: Context) => void;
 
+export type Headers = Record<string, string>;
+
 type Options = {
```
giorgiosironi, line 18: the name `Options` seem to have become too generic, even within a `context` module like this? Most of the keys seem HTTP- or routing-related but not all of them (`articles`) which are domain-specific instead.

test/create-article.ts
```diff
@@ -1,26 +1,31 @@
...
+} from 'rdf-js';
+import {
+  blankNode, dataset, literal, quad,
+} from '../src/data-factory';
+import { rdf, schema } from '../src/namespaces';
 
 type Options = {
-  id?: BlankNode;
-  name?: string;
+  id?: QuadSubject;
```
giorgiosironi, line 10: rightly more specific than `BlankNode`

test/middleware.ts
```diff
@@ -1,17 +1,15 @@
...
-): Promise<WithDataset<Response>> => {
+export default async <CustomT extends ExtendableContext>(
+  middleware: Middleware<CustomT>,
+  context: CustomT,
+  next?: Next<CustomT>,
+): Promise<CustomT['response']> => {
   await middleware(context, next ? (): Promise<void> => next(context) : jest.fn());
 
-  return context.response as WithDataset<Response>;
+  return context.response;
```
thewilkybarkid: Will extract these changes to a separate PR.

test/middleware/error-handler.test.ts
```diff
@@ -13,7 +13,7 @@ const makeRequest = async (
 ): Promise<WithDataset<Response>> => {
   const context = createContext({ errorListener });
 
-  return runMiddleware(errorHandler(), context, next);
+  return runMiddleware<ReturnType<typeof createContext>>(errorHandler(), context, next);
```
thewilkybarkid: Not sure why this is needed (see also https://github.com/libero/article-store/pull/115/commits/cae9366c72fcf4a7655b739332bb77523edf0ef6#r363793050).

thewilkybarkid: Worked it out, the type of `next`.

src/data-factory.ts
```diff
@@ -0,0 +1,11 @@
+import dataFactory from '@rdfjs/data-model';
+import createDataset from 'rdf-dataset-indexed';
+import { ExtendedDataFactory } from './middleware/dataset';
+
+const extendedDataFactory: ExtendedDataFactory = { ...dataFactory, dataset: createDataset };
```
giorgiosironi, line 5: a method name like `dataset()` could make it look like a singleton, whereas it's a factory that produces a new dataset each time it is called

test/middleware/error-handler.test.ts
```diff
@@ -36,12 +38,14 @@ describe('error-handler middleware', (): void => {
...
-      expect(object['@type']).toBe('http://www.w3.org/ns/hydra/core#Status');
-      expect(object['http://www.w3.org/ns/hydra/core#title']).toBe('Service Unavailable');
-      expect(object).not.toHaveProperty(['http://www.w3.org/ns/hydra/core#description']);
+      const id = blankNode();
+      const expected = [
+        quad(id, rdf.type, hydra.Status),
+        quad(id, hydra.title, literal('Service Unavailable', 'en')),
+      ];
+
+      expect([...response.dataset]).toEqualRdfQuadArray(expected);
```
giorgiosironi, line 46: asserting at an higher level than the JSON-LD body :+1:

test/middleware/jsonld.test.ts
```diff
@@ -1,105 +1,127 @@
...
-import runMiddleware, { Next } from '../middleware';
+import { rdf, schema } from '../../src/namespaces';
+import createContext, { Headers } from '../context';
+import { Next } from '../middleware';
 
-const makeRequest = async (next?: Next, jsonLdContext: JsonLdContext = {}): Promise<Response> => (
-  runMiddleware(jsonld(jsonLdContext), createContext(), next)
-);
+const makeRequest = async (
+  body?: string, headers?: Headers, next: Next = jest.fn(), jsonLdContext = {},
```
giorgiosironi: given the potential `context`/`jsonLdContext` confusion, could the latter have a better default, or type, to immediately make it clear what it is?

test/middleware/jsonld.test.ts
```diff
@@ -1,105 +1,127 @@
...
+const jsonLd = {
   '@id': 'http://example.com/object',
-  '@type': ['http://schema.org/Article'],
-  'http://schema.org/name': [{ '@value': 'English Name', '@language': 'en' }],
-  'http://purl.org/dc/elements/1.1/title': [
-    { '@value': 'English Title', '@language': 'en' },
-    { '@value': 'French Title', '@language': 'fr' },
-  ],
+  '@type': 'http://schema.org/Article',
+  'http://schema.org/name': [{ '@language': 'en', '@value': 'English Name' }, {
```
giorgiosironi: that's some weird formatting for two values that should be equally important (`en`, `fr`)

test/middleware/jsonld.test.ts
```diff
@@ -1,105 +1,127 @@
...
   });
 
-  it('should compact JSON-LD with a new context', async (): Promise<void> => {
+  it('does not override the response status code', async (): Promise<void> => {
+    const { response } = await makeRequest(null, null, next(null, quads, null, 201));
+
+    expect(response.status).toBe(201);
+  });
+
+  it('sets the response JSON-LD body with the dataset using a context', async (): Promise<void> => {
```
giorgiosironi, line 107: a JSON-LD context?

test/routes/add-article.test.ts
```diff
@@ -12,36 +14,42 @@ import createArticle from '../create-article';
...
 
   it('should throw an error if it is not a schema:Article', async (): Promise<void> => {
     const article = createArticle({ types: [schema.NewsArticle] });
 
     await expect(makeRequest(article)).rejects.toBeInstanceOf(createHttpError.BadRequest);
-    await expect(makeRequest(article)).rejects.toHaveProperty('message', 'Article type must be http://schema.org/Article (\'http://schema.org/NewsArticle\' was given)');
+    await expect(makeRequest(article)).rejects.toHaveProperty('message', 'No http://schema.org/Article found');
+  });
+
+  it('should throw an error if the articles does not have a blank node identifier', async (): Promise<void> => {
```
giorgiosironi, line 47: just an additional test or also an additional behavior? Keeps things simple but is there an hidden decision for not allowing a more complex identifier (like a DOI)?

giorgiosironi: (or maybe I'm confusing things)

src/middleware/dataset.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type ExtendedDataFactory = DataFactory & DatasetCoreFactory;
+
+export type DatasetContext<Context = DefaultContext> = DataFactoryContext<Context & {
+  request: WithDataset<Request>;
+  response: WithDataset<Response>;
+}, ExtendedDataFactory>;
+
+export default (): Middleware<DefaultState, DatasetContext> => (
+  async (context: DatasetContext, next: Next): Promise<void> => {
```
giorgiosironi: doesn't this mean that `context` passed in here already is a `DatasetContext`, hence it has the `.request.dataset` and `.response.dataset` fields; if so, who has added them before these `Object.assign`?

src/data-factory.ts
thewilkybarkid: It's https://rdf.js.org/dataset-spec/#datasetcorefactory-interface.

test/middleware.ts
thewilkybarkid: https://github.com/libero/article-store/pull/158

test/routes/add-article.test.ts
thewilkybarkid: For now an article has a blank node identifier (ie it isn't dereferenceable). In the near future it will have its own identifier, but it will still need to be generated by the server.

src/middleware/dataset.ts
thewilkybarkid: Think it was required when using `context.request = ...` with strict mode, but seems like it's ok with `Object.assign` so can remove.

thewilkybarkid: In fact, there's quite a few middleware where it's not quite right. Will push then extract.

thewilkybarkid: https://github.com/libero/article-store/pull/159

src/routes/add-article.ts
giorgiosironi: Good as-is following must forward.

test/middleware/routing.test.ts
```diff
@@ -17,7 +17,7 @@ const createRouter = (): Router => {
   return router;
 };
 
-const makeRequest = async (method: string, path: string, next?: Next): Promise<Response> => {
+const makeRequest = async (method: string, path: string, next?: NextMiddleware): Promise<Response> => {
```
thewilkybarkid: Spotted the wrong type is being used here. Will extract.

thewilkybarkid: https://github.com/libero/article-store/pull/161

test/context.ts
thewilkybarkid: Agree, but pre-existing so should change separately.

thewilkybarkid:
Think this is ready to :shipit:. (Finally! 😅)

giorgiosironi (approved):

PR #116, opened 2019-12-16
Replacing in-memory with persistent database storage

nlisgo:
Setup of postgres with docker. Exposes a script to create the articles database table. Creates a handler that implements `Articles` to store in and retrieves from the database.

- [x] Run the database table setup script on build
- [x] Set and override database connection details for different environments
- [x] Tests for the `persist-articles` handler
- [x] Unit tests can be run without database setup


.docker/docker-compose.yml
```diff
@@ -4,6 +4,8 @@ services:
   app:
     build:
       context: ../
+    environment:
+      DATABASE_CONNECTION_STRING: postgres://postgres:postgres@db:5432/articles
```
giorgiosironi: Reasonable to pass in environment variables for configuration. However, separating the components makes it easier to configure them as they could be managed differently:

- port hardcoded in a Helm chart (always `5432`)
- host generate by a Helm chart (e.g. `article-store--test--postgres`)
- password coming from a Kubernetes `Secret`

.docker/docker-compose.yml
```diff
@@ -15,3 +17,11 @@ services:
       - '8081:8000'
     depends_on:
       - app
+  db:
+    image: postgres:12.1-alpine
```
giorgiosironi: PostgreSQL 12 is not generally available yet on cloud providers such as [AWS RDS](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.DBVersions); stick to `11.5` at most

.docker/docker-compose.yml
```diff
@@ -15,3 +17,11 @@ services:
...
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: postgres
+      POSTGRES_DB: articles
```
giorgiosironi: the environment variables seem complete, can't think of anything else we need

.docker/docker-compose.yml
```diff
@@ -15,3 +17,11 @@ services:
...
+    ports:
+      - '5432:5432'
```
giorgiosironi, line 35: wouldn't use `volumes` yet

Makefile
```diff
@@ -33,6 +33,9 @@ start: ## Start the containers
 stop: ## Stop the containers
 	${DOCKER_COMPOSE} down
 
+initdb: ## Setup database tables
+	${DOCKER_COMPOSE} run app npm run initdb
```
giorgiosironi: `run --rm`?

.docker/docker-compose.yml
giorgiosironi: `articles` is also a table so a different name for `POSTGRES_DB` would help avoid confusion

giorgiosironi (commented):
Soft review to give direction to the Postgres setup, to make it compatible with Kubernetes deployments and managed databases such as RDS

Makefile
giorgiosironi: It's good for this to be available as an `npm` target in the image; this will probably be an `initContainer` in Kubernetes that is run before the Node.js application is started

src/adaptors/persist-articles.ts
```diff
@@ -0,0 +1,58 @@
+import { Iri, JsonLdObj, JsonLdArray } from 'jsonld/jsonld-spec';
+import { schema } from 'rdf-namespaces';
+import { IDatabase } from 'pg-promise';
+import uuidv4 from 'uuid/v4';
+import Articles from '../articles';
+import ArticleHasNoId from '../errors/article-has-no-id';
+import ArticleNotFound from '../errors/article-not-found';
+import NotAnArticle from '../errors/not-an-article';
+
+export default class PersistArticles implements Articles {
```
thewilkybarkid: `PostgresArticles` would be a clearer name.

src/adaptors/persist-articles.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    if (!(types.includes(schema.Article))) {
+      throw new NotAnArticle(types);
+    }
+
+    if (!('@id' in article)) {
+      throw new ArticleHasNoId();
+    }
+
+    this.db.none('INSERT INTO articles(uuid, article) VALUES ($[uuid], $[article])', {
```
thewilkybarkid:
```suggestion
    await this.db.none('INSERT INTO articles(uuid, article) VALUES ($[uuid], $[article])', {
```

src/adaptors/persist-articles.ts
```diff
@@ -0,0 +1,58 @@
...
+    const article = this.db.one('SELECT article FROM articles WHERE uuid = $[id]', { id }, (data: { article: JsonLdObj }) => data.article);
+    if (!article) {
+      throw new ArticleNotFound(id);
+    }
+
+    return article;
+  }
+
+  async remove(id: Iri): Promise<void> {
+    this.db.none('DELETE FROM articles WHERE uuid = $[id]', { id });
```
thewilkybarkid:
```suggestion
    await this.db.none('DELETE FROM articles WHERE uuid = $[id]', { id });
```

src/adaptors/persist-articles.ts
```diff
@@ -0,0 +1,58 @@
...
+import uuidv4 from 'uuid/v4';
+import Articles from '../articles';
+import ArticleHasNoId from '../errors/article-has-no-id';
+import ArticleNotFound from '../errors/article-not-found';
+import NotAnArticle from '../errors/not-an-article';
+
+export default class PersistArticles implements Articles {
+  private db: IDatabase<{}>;
+
+  public constructor(db: IDatabase<{}>) {
```
thewilkybarkid: Might want to use `IBaseProtocol`.

nlisgo: Have changed to `IBaseProtocol<IMain>`

nlisgo:
In order for the `.github/hypertest.sh` we need to ensure that the application is setup and the database tables have been created.

nlisgo:
`npm run test` is working for me locally. I suspect that this is because of the introduction of `pg-test`.

When I run the tests on my host machine this is the output before any tests are run:

```
➜  article-store git:(setup-postgres) npm run test

> @ test /Users/nathanlisgo/libero/article-store
> jest --silent

Determining test suites to run...circleci/postgres:10.6-alpine-ram already pulled (use ops.refreshImage to refresh)
Starting Docker Container pg-test
Waiting for pg-test on port 5433...
Waiting for pg-test on port 5433...
Setting pg connection string on environment: DATABASE_URL
Running pg migrations
```

giorgiosironi:
Will try to reproduce the build problems.

giorgiosironi:
Can reproduce the build with `make test`, but not sure what the error is about yet. `ENOENT` means a missing file or directory, but `jest` (for starters) is there in `node_modules/.bin`

nlisgo:
I'm trying locally to trigger the setup of postgres within the `./tests/adaptors/postgres-articles.test.ts`. It seems to be possible: https://www.atdatabases.org/docs/pg-test#api

I will push up what I have shortly.

nlisgo:
I have tests running locally successfully and the `pg-test` setup is just happening for `./tests/adaptors/postgres-articles.test.ts` rather than when any tests are run. There is some issues and redundancy in the setup. I have an issue where the migrations are supposed to be triggered in the `await pgSetup()` step but it doesn't seem to be working. I, therefore, trigger the creation of the `articles` table in the test setup. I then try to remove the configuration in `package.json` which references the pg test migration script but then my tests fail: :(

I'm still having the issue of only being able to run the tests successfully locally with `npm run test` but not with `make test`. See: https://github.com/libero/article-store/pull/116#issuecomment-572140838

test/adaptors/postgres-articles.test.ts
```diff
@@ -0,0 +1,147 @@
...
+import NotAnArticle from '../../src/errors/not-an-article';
+import { schema } from '../../src/namespaces';
+import createArticle from '../create-article';
+import CreateDb from '../../src/tools/create-db';
+
+let database: IBaseProtocol<IMain>;
+
+beforeAll(async (): Promise<void> => {
+  jest.setTimeout(10000);
+  database = await getDatabase().then((db) => pgPromise()(db.databaseURL));
```
giorgiosironi: I can see https://github.com/rrwen/pg-testdb/blob/master/index.js depends on https://github.com/olalonde/pgtools which has a JavaScript implementation of `createdb` and `dropdb` to manage logical database inside a Postgres server. This is within the realm of possibility, if performance allows (as the minimal teardown for tests or even test runs would be to clean the table as this `beforeEach` does, rather than using completely new databases).

I'm looking for signs of a new Postgres server being run, which would be out of place for application/test code to do; but I'm still digging.

giorgiosironi: Ah got the wrong one: https://github.com/ForbesLindesay/atdatabases/tree/master/packages/pg-test has a dependency on `@databases/with-container` which contains the [`docker` commands](https://github.com/ForbesLindesay/atdatabases/blob/master/packages/with-container/src/index.ts).

nlisgo:
It would be interesting to introduce groups in tests which is achievable with [jest-runner-groups](https://www.npmjs.com/package/jest-runner-groups). We could also put in place a `docker compose` override which doesn't have `db` as a dependency for tests which don't require it.

package.json
```diff
@@ -64,10 +68,13 @@
...
   "scripts": {
     "build": "tsc",
+    "initdb": "ts-node ./src/tools/initdb",
     "lint": "eslint . --ext .js,.ts",
     "lint:fix": "eslint . --fix --ext .js,.ts",
     "start": "node ./build/index.js",
     "start:dev": "nodemon --exec 'node --require ts-node/register ./src/index.ts' --ext .js,.json,.ts",
-    "test": "jest --silent"
+    "test": "jest --silent",
+    "test:unit": "jest --silent --testNamePattern '^((?!#integration).)*$'",
```
nlisgo: Since we are running all tests in CI we should not see any skipped tests there. The downside of running all tests in CI is that if we do not tag our integration tests with `#integration` in the describe name then it will still pass in CI.

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  async* [Symbol.asyncIterator](): AsyncIterator<[NamedNode, DatasetCore]> {
+    yield* await this.database.any('SELECT iri, article FROM articles').then((rows) => rows.map(async (row: { iri: string; article: string }) => [namedNode(row.iri), await fromStream(this.dataFactory.dataset(), this.parser.import(toReadableStream(row.article)))])) as [[NamedNode, DatasetCore]];
+  }
+
+  static async createTables(database: IBaseProtocol<IMain>): Promise<void> {
+    await database.none('DROP TABLE IF EXISTS articles');
+    await database.none(`CREATE TABLE IF NOT EXISTS articles (
+      iri varchar (128) NOT NULL UNIQUE,
```
nlisgo: 128 is just arbitrary. We need it to be long enough to accommodate an Iri.

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,83 @@
...
+      article text NOT NULL,
+      created time without time zone DEFAULT CURRENT_TIMESTAMP NOT NULL
```
nlisgo: We have no need for this yet so will remove it for now.

test/adaptors/postgres-articles.test.ts
```diff
@@ -0,0 +1,144 @@
...
+  await PostgresArticles.createTables(database);
+});
+
+afterAll((): void => {
+  pgp.end();
+});
+
+describe('postgres articles #integration', (): void => {
+  beforeEach(async (): Promise<void> => {
+    await database.none('TRUNCATE articles');
```
nlisgo: Other than setup this should have the same coverage as the `in-memory-articles.test.ts`: https://github.com/libero/article-store/blob/master/src/adaptors/in-memory-articles.ts

test/adaptors/postgres-articles.test.ts
nlisgo: We are no longer using this approach.

src/adaptors/postgres-articles.ts
giorgiosironi: Look for some de facto standard, there are usually limits in e.g. browsers that are on the same order of magnitude, hundreds or thousands of characters.

Since the first `I` in `IRI` means Internationalized, not sure how it plays out as a database usually need a physical measure and 128 may mean 128 bytes here rather than 128 actual characters.

src/tools/initdb.ts
```diff
@@ -0,0 +1,5 @@
+import pgPromise from 'pg-promise';
+import db from '../db';
+import PostgresArticles from '../adaptors/postgres-articles';
+
+PostgresArticles.createTables(pgPromise()(db));
```
thewilkybarkid: The promise here is ignored.

thewilkybarkid: See also https://stackoverflow.com/a/46515787.

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    return fromStream(this.dataFactory.dataset(), this.parser.import(toReadableStream(article)));
+  }
+
+  async remove(id: NamedNode): Promise<void> {
+    await this.database.none('DELETE FROM articles WHERE iri = $[iri]', { iri: id.value });
+  }
+
+  async contains(id: NamedNode): Promise<boolean> {
+    return this.database.one('SELECT COUNT(*) FROM articles WHERE iri = $[iri]', { iri: id.value }, (data: { count: number }) => +data.count > 0);
```
thewilkybarkid: Take a look at https://stackoverflow.com/a/16467634.

.docker/docker-compose.dev.yml
```diff
@@ -5,6 +5,12 @@ services:
     build:
       target: dev
     image: libero/article-store:${IMAGE_TAG:-master}-dev
+    environment:
+      DATABASE_NAME: article-store
+      DATABASE_USER: user
+      DATABASE_PASSWORD: secret
+      DATABASE_HOST: db
+      DATABASE_PORT: 5432
```
thewilkybarkid: I tend to prefer connection strings. @giorgiosironi?

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  async count(): Promise<number> {
+    return this.database.one('SELECT COUNT(*) FROM articles', [], (data: { count: number }) => +data.count);
```
thewilkybarkid: I see the tests fail if `+` is not there as `0` as a string is returned. By TypeScript thinks that `data.count` _is_ a `number`...

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  async* [Symbol.asyncIterator](): AsyncIterator<[NamedNode, DatasetCore]> {
+    yield* await this.database.any('SELECT iri, article FROM articles').then((rows) => rows.map(async (row: { iri: string; article: string }) => [namedNode(row.iri), await fromStream(this.dataFactory.dataset(), this.parser.import(toReadableStream(row.article)))])) as [[NamedNode, DatasetCore]];
```
thewilkybarkid: Might be one for later, but looks like `rows` is an array, so it must be loading everything _then_ mapping. It'd be better to stream/use a cursor to avoid problems on large tables.

.docker/docker-compose.dev.yml
nlisgo: > I tend to prefer connection strings. @giorgiosironi?

See: https://github.com/libero/article-store/pull/116#pullrequestreview-333301871

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  static async createTables(database: IBaseProtocol<IMain>): Promise<void> {
```
thewilkybarkid: Does more than just create tables. Maybe `setupTable`? `initializeTable`?

.docker/docker-compose.dev.yml
thewilkybarkid: :+1:

src/adaptors/postgres-articles.ts
thewilkybarkid: https://github.com/vitaly-t/pg-promise/wiki/BigInt

thewilkybarkid: Looked at various options, without switching `Articles.count()` to return `Promise<bigint>` this is the best option (but `count: number` needs to be `count: string`).

test/adaptors/postgres-articles.test.ts
thewilkybarkid: Unique table names is another option, and would allow tests to be run in parallel.

src/adaptors/postgres-articles.ts
nlisgo: Will not take this on now but it would be a valuable addition.

For reference: https://stackoverflow.com/questions/50694915/using-a-cursor-with-pg-promise/50698762#50698762

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,82 @@
...
+import SerializerJsonld from '@rdfjs/serializer-jsonld-ext';
+import pEvent from 'p-event';
+import { fromStream, toStream } from 'rdf-dataset-ext';
+import toReadableStream from 'to-readable-stream';
+import Articles from '../articles';
+import ArticleNotFound from '../errors/article-not-found';
+import NotAnArticle from '../errors/not-an-article';
+import { rdf, schema } from '../namespaces';
+import { ExtendedDataFactory } from '../middleware/dataset';
+import { namedNode } from '../data-factory';
```
thewilkybarkid: Should use the provided factory.

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export default class PostgresArticles implements Articles {
+  private database: IBaseProtocol<IMain>;
+
+  private dataFactory: ExtendedDataFactory;
+
+  private parser: Sink<Stream<Quad>, Stream<Quad>>;
+
+  private serialiser: Sink<Stream<Quad>, EventEmitter>;
```
thewilkybarkid: Bit simpler to just use `ParserJsonld` and `SerializerJsonld`.

thewilkybarkid: Probably better to call it `serializer`.

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  async* [Symbol.asyncIterator](): AsyncIterator<[NamedNode, DatasetCore]> {
+    yield* await this.database.any('SELECT iri, article FROM articles').then((rows) => rows.map(async (row: { iri: string; article: string }) => [namedNode(row.iri), await fromStream(this.dataFactory.dataset(), this.parser.import(toReadableStream(row.article)))])) as [[NamedNode, DatasetCore]];
+  }
+
+  static async setupTable(database: IBaseProtocol<IMain>): Promise<void> {
+    await database.none('DROP TABLE IF EXISTS articles');
+    await database.none(`CREATE TABLE IF NOT EXISTS articles (
+      iri varchar (128) NOT NULL UNIQUE,
+      article text NOT NULL
```
thewilkybarkid: Where did we get to with jsonb?

test/adaptors/postgres-articles.test.ts
```diff
@@ -0,0 +1,144 @@
...
+import pgPromise, { IBaseProtocol, IMain } from 'pg-promise';
+import PostgresArticles from '../../src/adaptors/postgres-articles';
+import ArticleNotFound from '../../src/errors/article-not-found';
+import NotAnArticle from '../../src/errors/not-an-article';
+import { schema } from '../../src/namespaces';
+import createArticle from '../create-article';
+import db from '../../src/db';
+import dataFactory from '../../src/data-factory';
+
+let pgp: IMain;
```
thewilkybarkid: Could use a clear name (thought of [something else](https://en.wikipedia.org/wiki/Pretty_Good_Privacy) on first glance!).

test/adaptors/postgres-articles.test.ts
thewilkybarkid: Similar to https://github.com/libero/content-api-bundle/blob/472c88159823a89bd71ed09534ab7234d7778b24/tests/Adapter/DoctrineItemsTest.php#L25-L57.

Makefile
```diff
@@ -69,18 +78,38 @@ fix: ## Fix linting issues in the code
...
+	$(MAKE) initdb
 	$(MAKE) run
 
 prod: export TARGET = prod
 prod: ## Builds and runs the container for production
 	$(MAKE) --jobs=2 build stop
+	$(MAKE) db
+	$(MAKE) wait-healthy
+	$(MAKE) initdb
 	$(MAKE) run
```
thewilkybarkid: Should be able to combine these into one (`$(MAKE) db wait-healthy initdb run`), as without `--jobs` they'll run in order.

Makefile
```diff
@@ -28,11 +28,20 @@ build: ## Build the containers
...
+	$(MAKE) initdb
 	${DOCKER_COMPOSE} up --detach
 
 stop: ## Stop the containers
 	${DOCKER_COMPOSE} down
 
+db: ## Start the database container
+	${DOCKER_COMPOSE} up --detach db
+
+initdb: ## Setup database tables
```
thewilkybarkid: The comments add this to the list of commands on `make help`, but we probably don't want that (so should be like `run`).

Makefile
thewilkybarkid: Though given the duplication, they should be part of `run` itself.

Makefile
```diff
@@ -69,18 +78,38 @@ fix: ## Fix linting issues in the code
...
+unit-test: export TARGET = dev
+unit-test: ## Run the unit tests
+	${DOCKER_COMPOSE} run --rm app npm run test:unit
+
+integration-test: export TARGET = dev
+integration-test: ## Run the integration tests
+	$(MAKE) db
+	$(MAKE) wait-healthy
+	$(MAKE) initdb
+	${DOCKER_COMPOSE} run --rm app npm run test:integration
```
thewilkybarkid: This will leave `db` running won't it?

Makefile
thewilkybarkid: And maybe combine `db`, `wait-healthy` and `initdb` into a single (private) command.

Makefile
```diff
@@ -69,18 +78,38 @@ fix: ## Fix linting issues in the code
...
 run:
 	${DOCKER_COMPOSE} up --abort-on-container-exit --exit-code-from app; ${DOCKER_COMPOSE} down
 
 dev: export TARGET = dev
 dev: ## Build and runs the container for development
 	$(MAKE) --jobs=4 install build stop
+	$(MAKE) db
+	$(MAKE) wait-healthy
+	$(MAKE) initdb
 	$(MAKE) run
```
thewilkybarkid: Related to https://github.com/libero/article-store/pull/116/files#r371827242, I guess the `docker-compose down` inside `run` does stop the `db` container?

test/adaptors/postgres-articles.test.ts
nlisgo: > Unique table names is another option, and would allow tests to be run in parallel.

This would be a good refactor. Should we take it on here or as a followup?

src/adaptors/postgres-articles.ts
nlisgo: > Where did we get to with jsonb?

I can try to reintroduce this.

test/adaptors/postgres-articles.test.ts
thewilkybarkid: It's ok for the moment, the only tests using the database are in this file, which are run in sequence.

test/adaptors/postgres-articles.test.ts
```diff
@@ -0,0 +1,143 @@
...
+  postgresPromise = pgPromise();
+  database = postgresPromise(db);
+});
+
+afterAll((): void => {
+  postgresPromise.end();
+});
+
+describe('postgres articles #integration', (): void => {
+  beforeEach(async (): Promise<void> => {
```
thewilkybarkid: I think this can exist outside the `describe`, in which case they're all together.

.docker/docker-compose.yml
```diff
@@ -4,6 +4,12 @@ services:
   app:
     build:
       context: ../
+    environment:
+      DATABASE_NAME: article-store
+      DATABASE_USER: user
+      DATABASE_PASSWORD: secret
+      DATABASE_HOST: db
+      DATABASE_PORT: 5432
```
giorgiosironi, line 12: saw a comment on connection strings. Advantages of multiple environment variables here:

- the more single environment variables can have sane defaults, the better to separate them rather than having a single string as it's easier to override just one or two of them. In the case of database connections though, the only variable with a default is `DATABASE_PORT`.
- the format of these variables is easier to get than a connection string (by memory: `postgres://user:pass@host:port/databasename`? Not sure if I got that right whereas I know I have with single variables)

Advantages of connection strings:

- if the application/library supports it, they can provide abstraction of the database driver as well e.g. passing in `sqlite://:memory:` rather than `postgres://...`
- fewer variables to document

For both cases, I wonder if there is any standard on the naming. This is Postgres-specific so the [`PG*` variable names](https://www.postgresql.org/docs/current/libpq-envars.html) could be used? Unless someone would think they will be magically picked up (like the `AWS_*` ones) rather than read by the application.

.github/hypertest.sh
```diff
@@ -8,5 +8,6 @@ function finish() {
 trap finish EXIT
 
 make start wait-healthy
+make initdb
```
giorgiosironi: shouldn't be necessary if it's in `make start`?

Makefile
giorgiosironi: Yes; can be unexpected. Catching errors, stopping it and raising the errors again seems an headache though.

Makefile
giorgiosironi: See no harm in pushing them in `run`.
Not sure if `$(MAKE)` propagates a `--jobs` down the recursive call?

giorgiosironi: Yes, asking for parallelism on the command line propagates:
```
$ cat Makefile 
.PHONY: first second

first:
        sleep 1
        echo first

second:
        sleep 1
        echo second

parent:
        $(MAKE) first second
$ time make --jobs=2 parent
make first second
make[1]: Entering directory '/home/giorgio/code/make-experiments'
sleep 1
sleep 1
echo first
echo second
first
second
make[1]: Leaving directory '/home/giorgio/code/make-experiments'

real    0m1.014s
user    0m0.006s
sys     0m0.012s
```

Makefile
thewilkybarkid: https://github.com/libero/article-store/blob/9f9e78331ae2c0e90d64d080b4c58667e44b19dc/Makefile#L76 works successfully.

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,81 @@
...
+  async* [Symbol.asyncIterator](): AsyncIterator<[NamedNode, DatasetCore]> {
+    yield* await this.database.any('SELECT iri, article FROM articles').then((rows) => rows.map(async (row: { iri: string; article: JsonLdObj }) => [this.dataFactory.namedNode(row.iri), await fromStream(this.dataFactory.dataset(), this.parser.import(toReadableStream(JSON.stringify(row.article))))])) as [[NamedNode, DatasetCore]];
+  }
+
+  static async setupTable(database: IBaseProtocol<IMain>): Promise<void> {
+    await database.none('DROP TABLE IF EXISTS articles');
+    await database.none(`CREATE TABLE IF NOT EXISTS articles (
+      iri varchar (128) NOT NULL UNIQUE,
+      article jsonb NOT NULL
+    )`);
```
thewilkybarkid: IIRC Postgres allows this to be inside a transaction.

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,81 @@
...
+    return this.database.one('SELECT COUNT(*) FROM articles', [], (data: { count: string }) => +data.count);
+  }
+
+  async* [Symbol.asyncIterator](): AsyncIterator<[NamedNode, DatasetCore]> {
+    yield* await this.database.any('SELECT iri, article FROM articles').then((rows) => rows.map(async (row: { iri: string; article: JsonLdObj }) => [this.dataFactory.namedNode(row.iri), await fromStream(this.dataFactory.dataset(), this.parser.import(toReadableStream(JSON.stringify(row.article))))])) as [[NamedNode, DatasetCore]];
+  }
+
+  static async setupTable(database: IBaseProtocol<IMain>): Promise<void> {
+    await database.none('DROP TABLE IF EXISTS articles');
+    await database.none(`CREATE TABLE IF NOT EXISTS articles (
```
thewilkybarkid: `IF NOT EXISTS` is redundant.

package.json
```diff
@@ -64,10 +67,13 @@
   },
   "scripts": {
     "build": "tsc",
+    "initdb": "ts-node ./src/tools/initdb",
```
thewilkybarkid: The current build failure is caused by this not working on the `prod` image, since it's no longer TypeScript.

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    return this.database.one('SELECT COUNT(*) FROM articles', [], callback);
+  }
+
+  async* [Symbol.asyncIterator](): AsyncIterator<[NamedNode, DatasetCore]> {
+    yield* await this.database.map('SELECT iri, article::text FROM articles ORDER BY id', [], this.toTuple, this);
+  }
+
+  static async setupTable(database: IBaseProtocol<IMain>): Promise<void> {
+    await database.tx(async (transaction: ITask<IMain>): Promise<void> => {
```
giorgiosironi, line 96: if this is surprising: https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis

nlisgo:
Thoughts on the approach for the `test/unit` and `test/integration` split. I want to extract the rename into a separate PR.

thewilkybarkid:
> Thoughts on the approach for the `test/unit` and `test/integration` split. I want to extract the rename into a separate PR.

I'm thinking that a second Jest config file is probably less invasive.  (And is then less of change if https://github.com/eugene-manuilov/jest-runner-groups/issues/4 works out.)

Makefile
```diff
@@ -69,10 +76,26 @@ fix: ## Fix linting issues in the code
...
+	${DOCKER_COMPOSE} run --rm app npm run test; ${DOCKER_COMPOSE} down
+
+unit-test: export TARGET = dev
+unit-test: ## Run the unit tests
+	${DOCKER_COMPOSE} run --rm app npm run test:unit
+
+integration-test: export TARGET = dev
+integration-test: ## Run the integration tests
+	$(MAKE) initdb
+	${DOCKER_COMPOSE} run --rm app npm run test:integration; ${DOCKER_COMPOSE} down
```
giorgiosironi: Doesn't this make the exit code the one from `docker-compose down`?

thewilkybarkid: Nope.

nlisgo: This test run should fail but without the exit code capture it just exits with a `0`

https://github.com/libero/article-store/runs/415379496#step:5:25

thewilkybarkid: Huh, locally I get a `2`.

thewilkybarkid: Oh no, I don't for that one... Tried something similar earlier and it worked? 🤔

Makefile
```diff
@@ -31,15 +31,19 @@ build: ## Build the containers
...
-	$(MAKE) initdb
+	$(MAKE) init-db
 	${DOCKER_COMPOSE} up --detach
 
+start-db:
+	${DOCKER_COMPOSE} up --detach db
+
+init-db:
+	$(MAKE) start-db
+	${DOCKER_COMPOSE} run --rm app npm run ${INITDB}
```
nlisgo, line 42: @thewilkybarkid Do we not need the `wait-healthy` anymore?

thewilkybarkid: It appears to make no difference. Hypothesis is that the database connection waits for it to be ready, but will see if we encounter a failure.

giorgiosironi: The driver probably has a timeout on the connection; one guarantee of the `postgres` container image is that when the port is opened, the database should be ready to work (they do an elaborate boostrap where they start listening only on localhost, run setup, then stop and restart the database server to listen to the outside world).

Makefile
```diff
@@ -69,10 +80,21 @@ fix: ## Fix linting issues in the code
 	${DOCKER_COMPOSE} run --rm app npm run lint:fix
 
 test: export TARGET = dev
-test: ## Run the tests
-	${DOCKER_COMPOSE} run --rm app npm run test
+test: ## Run all the tests
+	$(MAKE) start-db
+	${DOCKER_COMPOSE} run --rm app npm run test; EXIT=$$?; ${DOCKER_COMPOSE} down; exit $$EXIT
```
giorgiosironi: I forgot to change it after we introduced it, but `EXIT` should be lowercase as uppercase in shell is for environment variables rather than local variables?

Makefile
```diff
@@ -69,10 +80,21 @@ fix: ## Fix linting issues in the code
 	${DOCKER_COMPOSE} run --rm app npm run lint:fix
 
 test: export TARGET = dev
-test: ## Run the tests
-	${DOCKER_COMPOSE} run --rm app npm run test
+test: ## Run all the tests
+	$(MAKE) start-db
+	${DOCKER_COMPOSE} run --rm app npm run test; echo=$$?; ${DOCKER_COMPOSE} down; exit $$echp
```
giorgiosironi: `echp` typo?

src/db.ts
```diff
@@ -0,0 +1,11 @@
+import { IConnectionParameters } from 'pg-promise/typescript/pg-subset';
+
+const config: Readonly<IConnectionParameters> = {
+  user: process.env.DATABASE_USER,
+  password: process.env.DATABASE_PASSWORD,
+  database: process.env.DATABASE_NAME,
+  host: process.env.DATABASE_HOST,
+  port: process.env.DATABASE_PORT ? parseFloat(process.env.DATABASE_PORT) : undefined,
```
giorgiosironi: I understand there are no integers in JavaScript, but `parseInt` doesn't work? :thinking:

thewilkybarkid: Didn’t have time to work out what was going on with eslint. Understand now, but that rule applies to Browsers not Node so would need to disable. Doing the `+` trick elsewhere.

src/adaptors/postgres-articles.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  async count(): Promise<number> {
+    const callback = ({ count }: { count: string }): number => +count;
+
+    return this.database.one('SELECT COUNT(*) FROM articles', [], callback);
+  }
+
+  async* [Symbol.asyncIterator](): AsyncIterator<[NamedNode, DatasetCore]> {
+    yield* await this.database.map('SELECT iri, article::text FROM articles ORDER BY id', [], this.toTuple, this);
```
thewilkybarkid, line 92: Will need future work to scale, as it currently gets an array of everything then iterates through it.

thewilkybarkid (approved):

PR #121, opened 2019-12-17
Publish Docker image

giorgiosironi:
For https://github.com/libero/publisher/issues/310

The script should be later extracted into the `scripts` submodule.


giorgiosironi:
Added `DOCKER_*` secrets at https://github.com/libero/article-store/settings/secrets

.github/push-image.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
giorgiosironi: compare https://github.com/libero/scripts/blob/master/travis/push-image.sh

thewilkybarkid: We should move this there (`libero/script`).

If it needs testing, let's do it on a dummy repo.

.github/workflows/ci.yml
```diff
@@ -80,3 +80,27 @@ jobs:
 
       - name: Run test
         run: ${{ matrix.test }}
+
+  push:
+
+    needs: test
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
thewilkybarkid, line 89: Shame there's no simpler way.

giorgiosironi: The event types would limit to pushes, but not to a particular branch.

.github/push-image.sh
giorgiosironi: It should ideally tested in the `script` repository itself, to validate PRs.

giorgiosironi: Working on https://github.com/libero/scripts/pull/9 to eventually do that

giorgiosironi:
Compare https://github.com/libero/article-store/settings/secrets. Can't keep it up-to-date with `master` enough, but as ready as it can be.

thewilkybarkid (approved):
Looks like it should work.

PR #122, opened 2019-12-17
Add standard labels to prod container image

giorgiosironi:
Part of https://github.com/libero/publisher/issues/310

See https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys


giorgiosironi:
Wouldn't go overboard as adoption of the standard is unclear. Found the `revision` one useful in practice before while provisioning.

Real world example:
```
$ docker inspect elastic/elasticsearch:7.5.0 | jq .[].Config.Labels
{
  "org.opencontainers.image.created": "2019-11-26T01:06:52.520070Z",
  "org.opencontainers.image.documentation": "https://www.elastic.co/guide/en/elasticsearch/reference/index.html",
  "org.opencontainers.image.licenses": "Elastic-License",
  "org.opencontainers.image.revision": "e9ccaed468e2fac2275a3761849cbee64b39519f",
  "org.opencontainers.image.source": "https://github.com/elastic/elasticsearch",
  "org.opencontainers.image.title": "Elasticsearch",
  "org.opencontainers.image.url": "https://www.elastic.co/products/elasticsearch",
  "org.opencontainers.image.vendor": "Elastic",
  "org.opencontainers.image.version": "7.5.0"
}
```

Dockerfile
```diff
@@ -71,12 +71,16 @@ RUN npm run build
...
+ARG image_tag
 ENV NODE_ENV=production
 
 COPY --from=npm-prod /app/ .
 COPY --from=build-prod /app/build/ build/
 
 USER node
 
+LABEL org.opencontainers.image.revision=${image_tag} \
+      org.opencontainers.image.source=https://github.com/libero/article-store
```
thewilkybarkid: Been using 4-space indenting rather than alignment.

thewilkybarkid:
Feel like `licenses` might be a good idea.

Dockerfile
giorgiosironi: No opinions (but alignment in Terraform is terrible in fact)

giorgiosironi:
This should be `LABEL org.opencontainers.image.licenses=MIT` as the simplest possible SPDX license expression. However, the container image has [Node.js](https://github.com/nodejs/node/blob/master/LICENSE) and various other libraries inside it; this is probably why the label is in the plural form, but unclear how to list all the included licenses.

giorgiosironi:
A [general search](https://github.com/search?p=4&q=%22org.opencontainers.image.licenses%22&type=Code) indicates that people do not seem to be specifying multiple licenses ever. I Am Not A Lawyer so I will include MIT only.

thewilkybarkid:
We shouldn't be including anything that isn't compatible with an MIT license IIRC /cc @BlueReZZ.

giorgiosironi:
A discussion on licenses is needed in general if we are going to ship Docker images; however empirically it seems this label is only used to indicate the license of the core software present in the image rather than all the supporting components around it (not just libraries but also anything in `/bin`).

BlueReZZ:
Interesting that it's called `licences` plural. I may need to do some more research to get this right but my understanding is that the Docker image isn't licenced, only the software within it, and most OSS licences just require you to redistribute the licence with the software, so perhaps we need to know what licences are used and ensure they are alongside the code. I suppose they are analogous to CD-ROMs so we should find some advice online

thewilkybarkid:
😅

Let's take it out for now.

thewilkybarkid (approved):

PR #123, opened 2019-12-17
Refine CODEOWNERS to include Infrastructure Reviewers

giorgiosironi:
During https://github.com/libero/publisher/issues/310 [Node.js Reviewers](https://github.com/orgs/libero/teams/node-js-reviewers) were continuously requested for on scripts that aren't really related.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,5 +1,7 @@
 * @libero/node-js-reviewers
 /.docker/* @libero/infrastructure-reviewers
+/.dockerignore @libero/infrastructure-reviewers
 /.github/* @libero/infrastructure-reviewers
+/.scripts @libero/infrastructure-reviewers
 /Dockerfile @libero/infrastructure-reviewers
 /Makefile @libero/infrastructure-reviewers
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,7 +1,7 @@
 * @libero/node-js-reviewers
-/.docker/* @libero/infrastructure-reviewers
+/.docker/ @libero/infrastructure-reviewers
 /.dockerignore @libero/infrastructure-reviewers
-/.github/* @libero/infrastructure-reviewers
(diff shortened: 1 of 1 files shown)


giorgiosironi:
Though if Node.js reviewers are required anyway to merge, would need to request both?

thewilkybarkid:
Think it's ok. Could also have a merger group, but there's enough cross over anyway.

.github/CODEOWNERS
```diff
@@ -1 +1,3 @@
 * @libero/node-js-reviewers
+.github/* @libero/infrastructure-reviewers
+.docker/* @libero/infrastructure-reviewers
```
thewilkybarkid: Might want to include the `Dockerfile` and `Makefile`?

.github/CODEOWNERS
```diff
@@ -1 +1,5 @@
 * @libero/node-js-reviewers
+.docker/* @libero/infrastructure-reviewers
+.github/* @libero/infrastructure-reviewers
+Dockerfile @libero/infrastructure-reviewers
+Makefile @libero/infrastructure-reviewers
```
thewilkybarkid: Reading https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax these should be

```
/.docker/ @libero/infrastructure-reviewers
/.github/ @libero/infrastructure-reviewers
/Dockerfile @libero/infrastructure-reviewers
/Makefile @libero/infrastructure-reviewers
```

Also want `/.dockerignore`. `/.scripts` too? (Does that count as a file here? 🤔)

Not sure about `/.dependabot/`...

giorgiosironi: In theory `.scripts` should be a file containing the submodule commit so does count for modifications.

I'll go out on a limb and say `/.dependabot` is developer-oriented like the linting or testing configuration so shouldn't be included. It isn't very language-specific still.

thewilkybarkid (approved):

PR #128, opened 2019-12-19
Add operation types

thewilkybarkid:
Adds Schema.org operation types to the API.

Extracted from https://github.com/libero/article-store/pull/104.

### DIFF
--- a/src/routes/api-documentation.ts
+++ b/src/routes/api-documentation.ts
@@ -21,7 +21,7 @@ export default (): AppMiddleware => (
           '@type': hydra.Class,
           [hydra.title]: { '@value': 'API entry point', '@language': 'en' },
           [hydra.supportedOperation]: {
-            '@type': [hydra.Operation, schema.ViewAction],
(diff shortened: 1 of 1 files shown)


src/routes/api-documentation.ts
```diff
@@ -21,7 +21,7 @@ export default (): AppMiddleware => (
           '@type': hydra.Class,
           [hydra.title]: { '@value': 'API entry point', '@language': 'en' },
           [hydra.supportedOperation]: {
-            '@type': hydra.Operation,
+            '@type': [hydra.Operation, schema.ViewAction],
```
thewilkybarkid: Not sure if `ViewAction` is the most appropriate. `DownloadAction`? `ReceiveAction`?

thewilkybarkid: Though thinking maybe there's really no need to include `GET` operations, as Hydra resources _have_ to be dereferenceable?

asbjornu: What about [`ReadAction`](https://schema.org/ReadAction)?

PR #143, opened 2020-01-02
Change Articles.add to Articles.set

thewilkybarkid:
Makes it a bit more persistence-orientated, and might remove the need for a unit of work when we introduce a database.


giorgiosironi (approved):
Uncontroversial

PR #144, opened 2020-01-02
Remove RDF from the Hydra link header relation type

thewilkybarkid:
The value for the relation type is the same as `Hydra: apiDocumentation`, but isn't RDF so shouldn't use an RDF library.

Refs https://github.com/libero/article-store/pull/115#discussion_r357824304.

### DIFF
--- a/src/middleware/api-documentation-link.ts
+++ b/src/middleware/api-documentation-link.ts
@@ -1,6 +1,5 @@
 import formatLinkHeader from 'format-link-header';
 import { Context, Middleware, Next } from 'koa';
-import { hydra } from 'rdf-namespaces';
 import url from 'url';
 
 export default (path: string): Middleware => (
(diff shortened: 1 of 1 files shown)


src/middleware/api-documentation-link.ts
```diff
@@ -1,14 +1,13 @@
...
-import { hydra } from 'rdf-namespaces';
 import url from 'url';
 
 export default (path: string): Middleware => (
   async ({ request, response }: Context, next: Next): Promise<void> => {
     await next();
 
     const link = {
-      rel: hydra.apiDocumentation,
+      rel: 'http://www.w3.org/ns/hydra/core#apiDocumentation',
```
giorgiosironi, line 10: Clearly not RDF as it ends up into a `Link` header. Also not listed in https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types so assuming it's the Hydra specification dictating it.

giorgiosironi (approved):

PR #146, opened 2020-01-03
Use RDF/JS terms instead of JSON-LD strings

thewilkybarkid:
Currently strings are used for JSON-LD IRIs and blank node identifiers; the `Iri` type is just a string, so the two are being mixed. This introduces RDF/JS terms; while they still have to be converted to and from strings, it means the term type can be handled correctly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -453,6 +453,14 @@
         }
       }
     },
+    "@rdfjs/data-model": {
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/@rdfjs/data-model/-/data-model-1.1.2.tgz",
+      "integrity": "sha512-pk/G/JLYGaXesoBLvEmoC/ic0H3B79fTyS0Ujjh5YQB2DZW+mn05ZowFFv88rjB9jf7c1XE5XSmf8jzn6U0HHA==",
+      "requires": {
+        "@types/rdf-js": "^2.0.1"
+      }
+    },
     "@types/accepts": {
       "version": "1.3.5",
       "resolved": "https://registry.npmjs.org/@types/accepts/-/accepts-1.3.5.tgz",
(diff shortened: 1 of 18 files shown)


package-lock.json
```diff
@@ -723,8 +731,7 @@
     "@types/node": {
       "version": "12.12.21",
       "resolved": "https://registry.npmjs.org/@types/node/-/node-12.12.21.tgz",
-      "integrity": "sha512-8sRGhbpU+ck1n0PGAUgVrWrWdjSW2aqNeyC15W88GRsMpSwzv6RJGlLhE7s2RhVSOdyDmxbqlWSeThq4/7xqlA==",
-      "dev": true
```
thewilkybarkid, line 727: Not ideal. `@rdfjs/data-model` has a hard dependency on `@types/rdf-js` as it include its types. If it were a peer dependency, would that work? (`@types/node` is a local dev dependency.)

src/adaptors/in-memory-articles.ts
```diff
@@ -1,43 +1,45 @@
...
+  private articles: { [id: string]: [BlankNode, JsonLdObj] } = {};
 
-  async set(id: Iri, article: JsonLdObj): Promise<void> {
+  async set(id: BlankNode, article: JsonLdObj): Promise<void> {
     const types = [].concat(article['@type'] || []);
 
-    if (!(types.includes(schema.Article)) || article['@id'] !== id) {
-      throw new NotAnArticle(types);
+    if (!(types.includes(schema.Article)) || article['@id'] !== termToString(id)) {
+      throw new NotAnArticle(types.map((type: string): Term => stringToTerm(type)));
```
giorgiosironi, line 16: weird conversion due to the fact that `types` is not a RDF/JS object, but I guess it could go away later when the library is used more extensively

giorgiosironi (approved):
Consistent introduction of `BlankNode` and, to a lesser extent, of `Term`

src/adaptors/in-memory-articles.ts
thewilkybarkid: Yeah, it will.

package-lock.json
giorgiosironi: I guess peer dependencies would help making this dependency optional; but is it truly optional if the types are actually used? The intent of peer dependencies seems to me to specify compatibility without implying code inclusion.

PR #147, opened 2020-01-03
Create data factory middleware

thewilkybarkid:
Moves the data factory from the context prototype to a dedicated middleware with its own context, which will allow for more use of the context in future middleware.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -29,7 +29,6 @@ export default (
   const app = new Koa<AppState, AppContext>();
 
   app.context.articles = articles;
-  app.context.dataFactory = dataFactory;
   app.context.router = router;
 
   app.use(logger());
--- a/src/app.ts
+++ b/src/app.ts
@@ -6,6 +6,7 @@ import logger from 'koa-logger';
 import { DataFactory } from 'rdf-js';
 import Articles from './articles';
 import apiDocumentationLink from './middleware/api-documentation-link';
+import setDataFactory, { DataFactoryContext } from './middleware/data-factory';
 import emptyResponse from './middleware/empty-response';
 import errorHandler from './middleware/error-handler';
 import jsonld from './middleware/jsonld';
@@ -13,10 +14,9 @@ import routing from './middleware/routing';
 
 export type AppState = DefaultState;
 
-export type AppContext = RouterContext<AppState, {
+export type AppContext = RouterContext<AppState, DataFactoryContext<{
   articles: Articles;
-  dataFactory: DataFactory;
-}>;
+}>>;
 
 export type AppMiddleware = Middleware<AppState, AppContext>;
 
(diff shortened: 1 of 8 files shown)


giorgiosironi (approved):
The types don't seem like the best bang for the buck, but the refactoring is clear

thewilkybarkid:
Upped the typing game in https://github.com/libero/article-store/pull/115/commits/59ad69abcce8cf711ece32ebe0211199c9702528, will move that here.

src/middleware/data-factory.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  DefaultContext, DefaultState, Middleware, Next,
+} from 'koa';
+import { DataFactory } from 'rdf-js';
+
+export type DataFactoryContext<CustomT extends object = DefaultContext,
+  CustomF extends object = {}> = CustomT & { dataFactory: DataFactory & CustomF };
+
+export default <CustomF extends object>(dataFactory: DataFactory & CustomF):
```
thewilkybarkid: Trying to change the generic to `<CustomF extends DataFactory>`, but running in to a few problems in `data-factory.test.ts`.

src/app.ts
```diff
@@ -6,17 +6,17 @@ import logger from 'koa-logger';
...
+import setDataFactory, { DataFactoryContext } from './middleware/data-factory';
 import emptyResponse from './middleware/empty-response';
 import errorHandler from './middleware/error-handler';
 import jsonld from './middleware/jsonld';
 import routing from './middleware/routing';
 
 export type AppState = DefaultState;
 
-export type AppContext = RouterContext<AppState, {
+export type AppContext = RouterContext<AppState, DataFactoryContext<{
```
giorgiosironi, line 17: where's the type for `RouterContext`? DefinitelyTyped or somewhere else?

src/middleware/data-factory.ts
```diff
@@ -0,0 +1,16 @@
+import {
+  DefaultContext, DefaultState, Middleware, Next,
+} from 'koa';
+import { DataFactory } from 'rdf-js';
+
+export type DataFactoryContext<CustomT extends object = DefaultContext,
+  CustomF extends object = {}> = CustomT & { dataFactory: DataFactory & CustomF };
```
giorgiosironi: `T` and `F` don't sound very explanatory, they stand for `Type` and?

src/app.ts
thewilkybarkid: Yep https://github.com/libero/article-store/blob/dd041f613487e7b39f72ba3c3a4f3fcc8e89e3da/src/app.ts#L2

src/middleware/data-factory.ts
thewilkybarkid: `CustomT` using same name as Koa. `F` for factory, but these could be better.

src/app.ts
giorgiosironi: ah, I didn't see it in https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types because the list was too long

PR #151, opened 2020-01-06
Improve article factory options

thewilkybarkid:
Rather than manipulating the result, this adds more options to the article factory used in the tests.

Refs https://github.com/libero/article-store/pull/115#discussion_r358286937.

### DIFF
--- a/test/create-article.ts
+++ b/test/create-article.ts
@@ -1,14 +1,26 @@
+import { namedNode } from '@rdfjs/data-model';
 import deepFilter from 'deep-filter';
+import { Options } from 'jsonld';
 import { JsonLdObj } from 'jsonld/jsonld-spec';
-import { BlankNode } from 'rdf-js';
+import { BlankNode, NamedNode } from 'rdf-js';
 import { termToString } from 'rdf-string';
 
 const isNotUndefined = (arg: unknown): boolean => arg !== undefined;
 
-export default (id?: BlankNode, name = id ? `Article ${id}` : 'Article'): JsonLdObj => (
+type Options = {
+  id?: BlankNode;
+  name?: string;
+  types?: Array<NamedNode>;
+}
+
+export default ({
+  id = undefined,
+  name = 'Article',
+  types = [namedNode('http://schema.org/Article')],
+}: Options = {}): JsonLdObj => (
   deepFilter({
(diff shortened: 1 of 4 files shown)


test/create-article.ts
```diff
@@ -1,14 +1,26 @@
...
 import { termToString } from 'rdf-string';
 
 const isNotUndefined = (arg: unknown): boolean => arg !== undefined;
 
-export default (id?: BlankNode, name = id ? `Article ${id}` : 'Article'): JsonLdObj => (
+type Options = {
+  id?: BlankNode;
+  name?: string;
+  types?: Array<NamedNode>;
+}
```
thewilkybarkid, line 14: Using an object to avoid argument proliferation.

test/create-article.ts
```diff
@@ -1,14 +1,26 @@
...
 import { JsonLdObj } from 'jsonld/jsonld-spec';
-import { BlankNode } from 'rdf-js';
+import { BlankNode, NamedNode } from 'rdf-js';
 import { termToString } from 'rdf-string';
 
 const isNotUndefined = (arg: unknown): boolean => arg !== undefined;
 
-export default (id?: BlankNode, name = id ? `Article ${id}` : 'Article'): JsonLdObj => (
+type Options = {
+  id?: BlankNode;
```
giorgiosironi, line 11: I guess it doesn't make things easier to pass in a `string` `id`, as we need the `BlankNode` in the other calls to `set(...)` anyway.

test/create-article.ts
```diff
@@ -1,14 +1,26 @@
...
+
+export default ({
+  id = undefined,
+  name = 'Article',
+  types = [namedNode('http://schema.org/Article')],
+}: Options = {}): JsonLdObj => (
   deepFilter({
     '@id': id ? termToString(id) : undefined,
-    '@type': 'http://schema.org/Article',
+    '@type': types.map(termToString),
```
giorgiosironi, line 23: why passing in `NamedNode` instances if they have to be cast down to string anyway? Any validation happening in `namedNode()`?

thewilkybarkid: Yep, it needs to be an IRI.

giorgiosironi (approved):

PR #152, opened 2020-01-06
Add named node builder

thewilkybarkid:
Uses a builder rather than creating named nodes with an absolute IRI.

I've not used it in the JSON-LD usage as requires converting to a string, and this will be replaced in https://github.com/libero/article-store/pull/115.


src/namespaces.ts
```diff
@@ -0,0 +1,15 @@
+import namespace from '@rdfjs/namespace';
+
+const namespaces = {
```
giorgiosironi, line 3: `namespaces` is constant, but any way to make the values of the object constant as well? They can be reassigned and it is an exported variable :thinking:

giorgiosironi (approved):

PR #157, opened 2020-01-07
Fix issue with services loop in wait-healthy

nlisgo:
Currently, we are only checking the status of the first docker container. This ensures that all container ID's are passed to `./scripts/docker/wait-healthy.sh`.

~~Related and blocked by: https://github.com/libero/scripts/pull/15~~

If we can easily filter to only pass containers with healthchecks in place to `./scripts/docker/wait-healthy.sh` then we can close: https://github.com/libero/scripts/pull/15

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ wait-healthy: ## Wait for the containers to be healthy
 		echo "No containers running"; \
 		exit 1; \
 	fi; \
-	for service in $${services}; do \
+	for service in $${services[@]}; do \
 		.scripts/docker/wait-healthy.sh "$${service}"; \
 	done
 
--- a/.docker/docker-compose.yml
+++ b/.docker/docker-compose.yml
@@ -13,5 +13,9 @@ services:
       BASE_URL: http://localhost:8080/
     ports:
       - '8081:8000'
+    healthcheck:
+      test: wget --quiet --tries=1 --spider http://localhost:8000/ || exit 1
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -39,7 +39,7 @@ wait-healthy: ## Wait for the containers to be healthy
 		echo "No containers running"; \
 		exit 1; \
 	fi; \
-	for service in $${services}; do \
+	for service in $${services[@]}; do \
```
giorgiosironi, line 42: bugfix seems correct from quick experimentation, arcane Make syntax

giorgiosironi:
We should always have health checks in place to understand quickly if a container has a problem. Added one to the `console` service which was making the build fail.

thewilkybarkid:
We should add the healthcheck to the console image itself. /cc @tpluscode

giorgiosironi:
No reason not to push it upstream. Once it is in a release we can remove `healthcheck` from here and it will seamlessly pick up the `HEALTHCHECK` in the image.

giorgiosironi:
https://github.com/hypermedia-app/generic.hypermedia.app/pull/50

Makefile
thewilkybarkid: Never seen it before, but https://stackabuse.com/array-loops-in-bash/#loopingoverarrays suggests it's Bash?

thewilkybarkid (approved):

PR #158, opened 2020-01-07
Improve request and response typing in tests

thewilkybarkid:
Fixes the request and response types in the context factory, which allows the middleware runner to return the real response type. 

Refs https://github.com/libero/article-store/pull/115#discussion_r363793050.

### DIFF
--- a/test/middleware.ts
+++ b/test/middleware.ts
@@ -1,18 +1,16 @@
-import { Middleware, ParameterizedContext, Response } from 'koa';
-import { AppContext, AppState } from '../src/app';
+import { ExtendableContext } from 'koa';
+import { Middleware } from 'koa-compose';
+import { AppContext } from '../src/app';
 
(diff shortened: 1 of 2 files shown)


giorgiosironi (approved):

test/middleware.ts
```diff
@@ -1,18 +1,16 @@
...
-export default async <StateT = AppState, CustomT = AppContext>(
-  middleware: Middleware<StateT, CustomT>,
-  context: ParameterizedContext<StateT, CustomT>,
-  next?: Next<StateT, CustomT>,
-): Promise<Response> => {
+export default async <CustomT extends ExtendableContext>(
+  middleware: Middleware<CustomT>,
+  context: CustomT,
+  next?: Next<CustomT>,
+): Promise<CustomT['response']> => {
```
giorgiosironi, line 12: first time I see this notation to navigate to the type of a field, but it's understandable

PR #161, opened 2020-01-08
Fix next middleware type

thewilkybarkid:
Some tests are using the wrong type for the middleware runner, this renames it so it's different from Koa's `Next` type and fixes its usage.

### DIFF
--- a/test/middleware.ts
+++ b/test/middleware.ts
@@ -1,14 +1,14 @@
 import { Middleware, ParameterizedContext, Response } from 'koa';
 import { AppContext, AppState } from '../src/app';
 
-export type Next<StateT = AppState, CustomT = AppContext> = (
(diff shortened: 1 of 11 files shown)


test/middleware.ts
```diff
@@ -2,13 +2,13 @@ import { ExtendableContext } from 'koa';
 import { Middleware } from 'koa-compose';
 import { AppContext } from '../src/app';
 
-export type Next<CustomT extends ExtendableContext = AppContext> =
+export type NextMiddleware<CustomT extends ExtendableContext = AppContext> =
   (context: CustomT) => Promise<void>;
```
giorgiosironi, line 6: comparison with https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/koa/index.d.ts#L736 which has no arguments

giorgiosironi (approved):

PR #164, opened 2020-01-09
Correct types in tests

thewilkybarkid:
Resolves some type issues that are revealed when strict mode is enabled. (https://github.com/libero/article-store/pull/29)

### DIFF
--- a/test/create-article.ts
+++ b/test/create-article.ts
@@ -8,7 +8,7 @@ import { rdf, schema } from '../src/namespaces';
 
 type Options = {
   id?: QuadSubject;
-  name?: Literal;
+  name?: Literal | null;
   types?: Array<NamedNode>;
 }
 
--- a/test/middleware/jsonld.test.ts
+++ b/test/middleware/jsonld.test.ts
@@ -21,7 +21,7 @@ const makeRequest = async (
   return context;
 };
 
-const next = (body = null, quads: Array<Quad> = [], type = null, status = null) => (
+const next = (body?: unknown, quads?: Array<Quad>, type?: string, status?: number) => (
   async ({ response }: AppContext): Promise<void> => {
     if (body) {
       response.body = body;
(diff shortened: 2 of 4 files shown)


test/middleware/jsonld.test.ts
```diff
@@ -72,7 +72,7 @@ describe('JSON-LD middleware', (): void => {
   });
 
   it('sets the response JSON-LD body with the dataset', async (): Promise<void> => {
-    const { response } = await makeRequest(null, null, next(null, quads));
+    const { response } = await makeRequest(undefined, undefined, next(undefined, quads));
```
giorgiosironi, line 75: not that readable in terms of parameters, but didn't change from before

test/create-article.ts
```diff
@@ -8,7 +8,7 @@ import { rdf, schema } from '../src/namespaces';
 
 type Options = {
   id?: QuadSubject;
-  name?: Literal;
+  name?: Literal | null;
```
giorgiosironi, line 11: only example that needs `null` in addition to the `undefined` provided by `?`?

giorgiosironi (approved):

test/create-article.ts
thewilkybarkid: https://github.com/libero/article-store/blob/a6762f995cdd3c277786b858e789a17282b2790a/test/routes/add-article.test.ts#L55-L56

test/middleware/jsonld.test.ts
thewilkybarkid: Other option is to use an object, but is a wider refactor.

giorgiosironi: (unrelated here I agree)

test/create-article.ts
giorgiosironi: So it's by choice, but it's an error case so wouldn't expect it to mean something in particular like "a title that is currently unknown"

PR #165, opened 2020-01-09
Don't used untyped modules

thewilkybarkid:
Removes usage in the tests of modules that are not typed, which should not be considered public.

Refs https://github.com/libero/article-store/pull/29#issuecomment-572475448.

### DIFF
--- a/test/router.test.ts
+++ b/test/router.test.ts
@@ -1,4 +1,3 @@
-import Layer from '@koa/router/lib/layer';
 import createRouter from '../src/router';
 import Routes from '../src/routes';
 
@@ -7,7 +6,7 @@ describe('router', (): void => {
 
   Object.keys(Routes).forEach((name: string): void => {
     it(`has a route named ${name}`, (): void => {
-      expect(router.route(Routes[name])).toBeInstanceOf(Layer);
+      expect(router.route(Routes[name])).toBeTruthy();
     });
   });
 });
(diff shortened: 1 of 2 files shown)


test/router.test.ts
```diff
@@ -7,7 +6,7 @@ describe('router', (): void => {
 
   Object.keys(Routes).forEach((name: string): void => {
     it(`has a route named ${name}`, (): void => {
-      expect(router.route(Routes[name])).toBeInstanceOf(Layer);
+      expect(router.route(Routes[name])).toBeTruthy();
```
thewilkybarkid: No way to access the object unfortunately. `route()` returns a `Layer` or `false`.

giorgiosironi (approved):
Though dropping them because they are private, more than because of the lack of typing?

test/router.test.ts
giorgiosironi: to be an object (stronger than being truthy)?

giorgiosironi: (no idea of the matchers available here)

thewilkybarkid:
More because of the latter, but wasn't exactly happen with using them.

test/router.test.ts
thewilkybarkid: https://github.com/libero/article-store/pull/165/commits/92f126a3363e636917734b11238e2d190cca4969 is the best I can think of that still gives a useful message.

test/router.test.ts
```diff
@@ -7,7 +6,7 @@ describe('router', (): void => {
 
   Object.values(Routes).forEach((name: string): void => {
     it(`has a route named ${name}`, (): void => {
-      expect(router.route(name)).toBeInstanceOf(Layer);
+      expect(router.route(name)).toMatchObject({ name });
```
giorgiosironi, line 9: even better than a simple `object` type

PR #166, opened 2020-01-09
Fix router context

thewilkybarkid:
The Router can't know about `RouterContext`; this replaces uses its contents instead.

This becomes visible with script mode turned on.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -14,11 +14,11 @@ import namespaces from './namespaces';
 
 export type AppState = DefaultState;
 
-export type AppRouterContext = DatasetContext<{
+export type AppServiceContext = DatasetContext<{
   articles: Articles;
 }>;
 
-export type AppContext = RouterContext<AppState, AppRouterContext>;
+export type AppContext = RouterContext<AppState, AppServiceContext>;
 
 export type AppMiddleware = Middleware<AppState, AppContext>;
 
@@ -26,7 +26,7 @@ export type App = Koa<AppState, AppContext>
 
 export default (
   articles: Articles,
-  router: Router<AppState, AppRouterContext>,
+  router: Router<AppState, AppServiceContext>,
   apiDocumentationPath: string,
   dataFactory: ExtendedDataFactory,
 ): App => {
--- a/src/app.ts
+++ b/src/app.ts
@@ -14,9 +14,11 @@ import namespaces from './namespaces';
 
 export type AppState = DefaultState;
 
-export type AppContext = RouterContext<AppState, DatasetContext<{
+export type AppRouterContext = DatasetContext<{
   articles: Articles;
-}>>;
+}>;
+
+export type AppContext = RouterContext<AppState, AppRouterContext>;
 
 export type AppMiddleware = Middleware<AppState, AppContext>;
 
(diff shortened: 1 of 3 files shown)


src/middleware/routing.ts
```diff
@@ -1,22 +1,26 @@
...
+import {
+  DefaultContextExtends, DefaultStateExtends, Middleware, Next,
+} from 'koa';
 import compose from 'koa-compose';
 
-const notFound = (): Middleware<DefaultState, RouterContext> => (
-  async ({ _matchedRoute }: RouterContext, next: Next): Promise<void> => {
-    await next();
+const notFound = <State extends DefaultStateExtends, Context extends DefaultContextExtends>
+  (): Middleware<State, RouterContext<State, Context>> => (
```
thewilkybarkid: This indentation is annoying.

src/app.ts
```diff
@@ -14,17 +14,19 @@ import namespaces from './namespaces';
 
 export type AppState = DefaultState;
 
-export type AppContext = RouterContext<AppState, DatasetContext<{
+export type AppRouterContext = DatasetContext<{
```
thewilkybarkid: Is there a better name for this?

giorgiosironi: Could keep using the `Custom` prefix in the name?

giorgiosironi: (from `RouterContext<StateT = any, CustomT = {}>`)

thewilkybarkid:
Just seen a failing healthcheck for the console for the first time (https://github.com/libero/article-store/runs/381516086). A re-run sorted it (https://github.com/libero/article-store/runs/381564521). No other details unfortunately (nothing logged), but something to keep an eye on. /cc @tpluscode

src/app.ts
thewilkybarkid: Hmm, `CustomContext` doesn't sound great...

giorgiosironi: Looking at it another way, this is extending the (already extended) context with `articles` and possibly other dependencies; assuming it's because they need to be injected into the routes, so this has become a Service Locator / DI Container...

thewilkybarkid: Good idea, let's go for `ServiceContext`.

src/middleware/routing.ts
```diff
@@ -1,22 +1,24 @@
...
-import { Middleware, Next, DefaultState } from 'koa';
+import { DefaultContextExtends, DefaultStateExtends, Next } from 'koa';
 import compose from 'koa-compose';
 
-const notFound = (): Middleware<DefaultState, RouterContext> => (
-  async ({ _matchedRoute }: RouterContext, next: Next): Promise<void> => {
-    await next();
+const notFound = <State extends DefaultStateExtends, Context extends DefaultContextExtends>
+  (): Middleware<State, Context> => (
+    async ({ _matchedRoute }: RouterParamContext<State, Context>, next: Next): Promise<void> => {
```
giorgiosironi, line 8: yes, this is weird indentation; it changes depending on how the previous line is broken up?

giorgiosironi (approved):

src/middleware/routing.ts
thewilkybarkid: Seems to be a difference between an assignment and an export...

PR #167, opened 2020-01-09
Set up mutation tests

thewilkybarkid:
Configures and runs [Stryker](https://stryker-mutator.io/), a mutation tester.

PR #172, opened 2020-01-10
Fix dataset middleware test

thewilkybarkid:
Currently the test can't actually check that the dataset is set on the request and response as the context factory has already added one.

I'm not entirely happy with this initial solution so leaving it in draft for now.

Found by https://github.com/libero/article-store/pull/167.


thewilkybarkid:
Closing in favour of #271.

PR #173, opened 2020-01-10
Jest runner groups

nlisgo:
We need to introduce groups in our tests so that unit tests can be run with minimal setup while integration tests can be trigger after additional steps for setup.

I'm introducing here to reduce the diff of introducing it into: https://github.com/libero/article-store/pull/116

### DIFF
--- a/test/adaptors/in-memory-articles.test.ts
+++ b/test/adaptors/in-memory-articles.test.ts
@@ -8,6 +8,9 @@ import NotAnArticle from '../../src/errors/not-an-article';
 import { schema } from '../../src/namespaces';
 import createArticle from '../create-article';
 
+/**
+ * @group unit
+ */
 describe('in-memory articles', (): void => {
   it('can add an article', async (): Promise<void> => {
     const articles = new InMemoryArticles();
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,5 +1,6 @@
 module.exports = {
   rootDir: 'test',
+  runner: 'groups',
   transform: {
     '^.+\\.ts$': 'ts-jest',
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -4815,6 +4815,16 @@
         "throat": "^4.0.0"
       }
     },
+    "jest-runner-groups": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/jest-runner-groups/-/jest-runner-groups-1.0.0.tgz",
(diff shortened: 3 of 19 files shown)


test/adaptors/in-memory-articles.test.ts
```diff
@@ -8,6 +8,9 @@ import NotAnArticle from '../../src/errors/not-an-article';
 import { schema } from '../../src/namespaces';
 import createArticle from '../create-article';
 
+/**
+ * @group unit
```
thewilkybarkid, line 12: Any way to have a default group, or exclude groups when running? Would rather just tag certain ones.

nlisgo: > Any way to have a default group, or exclude groups when running? Would rather just tag certain ones.

I will look into this. It would be an improvement.

giorgiosironi:
For now just saying, good to implement this independently of database-related changes.

nlisgo:
A couple of alternative options is to clearly distinguish the integration tests in the name of the test files are in the `describe` `name` parameter and then we can apply a filter when the tests are run.

In an implementation for the work in introducing database persistence I add a tag to the `describe` `name` parameter (https://github.com/libero/article-store/pull/116/files#diff-6a9286bba3afe3109741264f306a3655R27) and then apply a filter in running the `unit` and `integration` tests (https://github.com/libero/article-store/pull/116/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R75-R76).

A weakness of that approach is that in each test run the tests that are filtered out appear as `skipped` which is not ideal. I would prefer that no tests appear skipped.

giorgiosironi:
For consistency, I would specify 3 different Makefile targets:

- `make test`: runs everything (hence has to start the database)
- `make unit-test`: runs all tests with no #integration tag (no database)
- `make integration-test`: runs all test with the #integration tag (has to start the database too)

The `docker-compose*.yml` duplication is annoying but don't see an easy way to remove it without breaking up everything into even more YAML files to merge.

nlisgo:
I'm closing this. I have implemented a different approach when introducing the first integration tests in https://github.com/libero/article-store/pull/116

We can discuss the approach there and potentially develop a smaller PR with only those changes.

test/adaptors/in-memory-articles.test.ts
thewilkybarkid: Opened https://github.com/eugene-manuilov/jest-runner-groups/issues/4#issue-556885899.

PR #181, opened 2020-01-14
Change blanknodes to namednodes

nlisgo:
The aim of this PR is to introduce the smallest amount of changes to swap out `BlankNodes` for `NamedNodes`.

### DIFF
--- a/src/routes/add-article.ts
+++ b/src/routes/add-article.ts
@@ -12,7 +12,7 @@ import Routes from './index';
 
 export default (): AppMiddleware => (
   async ({
-    articles, dataFactory: { blankNode, quad }, request, response, router,
+    articles, dataFactory: { namedNode, quad }, request, response, router,
   }: AppContext, next: Next): Promise<void> => {
     const id = clownface({ dataset: request.dataset }).has(rdf.type, schema.Article).term;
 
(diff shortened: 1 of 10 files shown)


nlisgo:
@giorgiosironi I would like this to be considered for release. I know it doesn't address all of the todo items but it adds enough value in my opinion. The next piece of work will be to improve the identifier to include the `/articles/` prefix and then we will want to allow the retrieval of a single article in the store. 

I believe this will be enough to unblock the database persistence work.

src/routes/add-article.ts
```diff
@@ -28,7 +28,7 @@ export default (): AppMiddleware => (
       throw new createHttpError.BadRequest(`Article must have at least one ${termToString(schema('name'))}`);
     }
 
-    const newId = blankNode(uniqueString());
+    const newId = namedNode(uniqueString());
```
giorgiosironi, line 31: The documentation says [this should be an IRI](http://rdf.js.org/data-model-spec/#namednode-interface), but we have [good reasons to use host-free values](
https://github.com/libero/publisher/issues/367#issuecomment-574638754).

giorgiosironi (approved):
Checking NamedNode basic requirements I think this should go ahead.
The other tasks in https://github.com/libero/publisher/issues/367 however still need to be followed up before persistence can be introduced as the whole ticket has been created as a blocker (also I see that at the bare minimum identifiers like `/articles/...` should be stable before they start to get persisted).

giorgiosironi:
Found out this couldn't be merged by @nlisgo because the [branch protection on `master`](https://github.com/libero/article-store/settings/branch_protection_rules/11906254) limits pushing to `Organization administrators, repository administrators, and users with the Maintain role` plus Dependabot. https://github.com/orgs/libero/teams/node-js-reviewers only has `Write`, not `Maintain` permission.

PR #184, opened 2020-01-16
Uses Iri as NamedNode value and allow individual article retrieval

nlisgo:
This is a followup to: https://github.com/libero/article-store/pull/181

It uses an Iri as the NamedNode value and allwos the retrieval of an individual article.


giorgiosironi:
Screenshot_2020-01-16 Uses Iri as NamedNode value and allow individual article retrieval by nlisgo · Pull Request #184 · li

If someone edits the description of an issue you have created, your spirit will be evoked.

giorgiosironi:
Now that merge of https://github.com/libero/article-store/pull/181 is possible, should be rebased to remove the changes that are already in.

src/routes/article.ts
```diff
@@ -0,0 +1,31 @@
...
+      const graph = clownface({
+        dataset: response.dataset,
+        term: articleNamedNode,
+      });
+      try {
+        await articles.get(articleNamedNode).then((article) => {
+          addAll(graph.dataset, article);
+        });
+      } catch (error) {
+        throw new createHttpError.NotFound(error.message);
```
thewilkybarkid: The `try`/`catch` block shouldn't wrap other code (only the `get`), and it should check for the error type (`ArticleNotFound`).  At the moment _any_ error, including those from Clownface are turned into `NotFound`.

src/router.ts
```diff
@@ -13,6 +14,8 @@ export default (): Router<AppState, AppServiceContext> => {
   router.get(Routes.ArticleList, '/articles', articleList());
   router.post(Routes.AddArticle, '/articles', addArticle());
   router.get(Routes.EntryPoint, '/', entryPoint());
+  router.get(Routes.Article, '/articles/:id', article());
```
thewilkybarkid, line 16: We can't use the router for individual articles, as the values are persistent (the 'route' could change, but the existing values won't).

test/context.ts
```diff
@@ -39,6 +40,7 @@ export default ({
   method,
   path,
   router = dummyRouter,
+  url,
```
thewilkybarkid: Looks like it's the same as `path`?

src/routes/article.ts
```diff
@@ -0,0 +1,42 @@
...
+import { namedNode } from '../data-factory';
+import ArticleNotFound from '../errors/article-not-found';
+import { DatasetContext } from '../middleware/dataset';
+
+export default (articles: Articles): Middleware<DefaultStateExtends, DatasetContext> => (
+  async ({ request, response }: DatasetContext, next: Next): Promise<void> => {
+    try {
+      await next();
+    } catch (error) {
+      if (error instanceof createHttpError.NotFound) {
```
thewilkybarkid: Should be able to return early from both a successful `next()` and an error that isn't a `NotFound`.

src/routes/article.ts
```diff
@@ -0,0 +1,42 @@
...
+            throw new createHttpError.NotFound(getError.message);
+          }
+
+          throw getError;
+        }
+        const graph = clownface({
+          dataset: response.dataset,
+          term: articleNamedNode,
+        });
+        addAll(graph.dataset, article);
```
thewilkybarkid: `response.dataset = article`?

src/app.ts
```diff
@@ -48,6 +49,7 @@ export default (
   }));
   app.use(apiDocumentationLink(apiDocumentationPath));
   app.use(errorHandler());
+  app.use(article(articles));
```
thewilkybarkid: Did this not work as a router middleware in the end?

thewilkybarkid: Just experimented and it looks like router middleware aren't called if the route isn't found. Which makes sense.

thewilkybarkid: Just seen https://github.com/koajs/router/issues/44.

src/routes/article.ts
```diff
@@ -0,0 +1,42 @@
...
+  DefaultStateExtends, Middleware, Next,
+} from 'koa';
+import url from 'url';
+import Articles from '../articles';
+import { namedNode } from '../data-factory';
+import ArticleNotFound from '../errors/article-not-found';
+import { DatasetContext } from '../middleware/dataset';
+
+export default (articles: Articles): Middleware<DefaultStateExtends, DatasetContext> => (
+  async ({ request, response }: DatasetContext, next: Next): Promise<void> => {
```
thewilkybarkid: (Assuming this doesn't become a router middleware) Use `AppState` and `AppServiceContext`, and so `articles` from the context.

test/routes/article.test.ts
```diff
@@ -0,0 +1,83 @@
...
+import InMemoryArticles from '../../src/adaptors/in-memory-articles';
+import Articles from '../../src/articles';
+import article from '../../src/routes/article';
+import createContext from '../context';
+import createArticle from '../create-article';
+import runMiddleware, { NextMiddleware } from '../middleware';
+import { WithDataset } from '../../src/middleware/dataset';
+import ArticleNotFound from '../../src/errors/article-not-found';
+
+const inMemoryArticles = new InMemoryArticles();
```
thewilkybarkid: This will be shared between test cases, so could lead to unexpected results. Each test should have its own.

test/routes/article.test.ts
```diff
@@ -0,0 +1,83 @@
...
+      count: jest.fn(),
+      [Symbol.asyncIterator]: jest.fn(),
+    };
+
+    const next = jest.fn();
+    await makeRequest('path-to/article/one', mockArticles, next);
+    expect(mockArticles.get).toHaveBeenCalledTimes(0);
+    await expect(makeRequest('path-to/article/one', mockArticles)).rejects.toBeInstanceOf(createHttpError.NotFound);
+    expect(mockArticles.get).toHaveBeenCalledTimes(1);
+    expect(mockArticles.get).toHaveBeenCalledWith(namedNode('http://example.com/path-to/article/one'));
```
thewilkybarkid: These are two different test cases.

test/routes/article.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  it('should throw an error if article is not found', async (): Promise<void> => {
+    const response = makeRequest('path-to/article/not-found');
+
+    await expect(response).rejects.toBeInstanceOf(createHttpError.NotFound);
+    await expect(response).rejects.toHaveProperty('message', 'Article http://example.com/path-to/article/not-found could not be found');
```
thewilkybarkid: Better to use `.toHaveProperty('id', id)`, since a wording change inside the error constructor would break this.

src/routes/article.ts
thewilkybarkid: Think might be able to pass this as a 'not found' handler to the routing middleware, so this middleware itself doesn't have to handle it.

nlisgo: > Think might be able to pass this as a 'not found' handler to the routing middleware, so this middleware itself doesn't have to handle it.

Isn't the routing middleware run before the get article request is handled. How will the 'not found' handler be passed to the routing middleware unless we amend that workflow?

src/routes/article.ts
nlisgo: We have the issue of `Dependency cycle detected.`

nlisgo: We would need `AppContext` to access `path` from the context.

test/context.ts
```diff
@@ -22,6 +22,7 @@ type Options = {
   method?: string;
   path?: string;
   router?: Router;
+  url?: string;
```
thewilkybarkid:
```suggestion
```

src/routes/article.ts
```diff
@@ -0,0 +1,35 @@
...
+import { OK } from 'http-status-codes';
+import {
+  DefaultStateExtends, Middleware, Next,
+} from 'koa';
+import url from 'url';
+import { AppContext } from '../app';
+import { namedNode } from '../data-factory';
+import ArticleNotFound from '../errors/article-not-found';
+
+export default (): Middleware<DefaultStateExtends, AppContext> => (
```
thewilkybarkid: `AppMiddleware`

src/routes/article.ts
```diff
@@ -0,0 +1,35 @@
...
+    path, articles, request, response,
+  }: AppContext, next: Next): Promise<void> => {
+    try {
+      await next();
+    } catch (error) {
+      if (!(error instanceof createHttpError.NotFound)) {
+        throw error;
+      }
+
+      try {
```
thewilkybarkid: If you add a `return` in the end of the `try` and close the `catch` straight after the `if` statement, we can avoid nesting `try` blocks.

src/routes/article.ts
nlisgo: Happy to do this but we're still left with the `Dependency cycle detected` lint error.

thewilkybarkid:
I've pushed https://github.com/libero/article-store/pull/184/commits/0d43a9a3266b180f31fc88f16e17f8d4c267191c which does contradict https://github.com/libero/article-store/pull/184#discussion_r368580520. There's complexity in the article 'route' having to know whether to lookup or not (could be extracted though), and the dependency cycle (https://github.com/libero/article-store/pull/184#discussion_r370007860; again could be solved by extracting).

Not 100% happy with this, means having to add another route if the format is ever changed. Will continue to think.

nlisgo:
> Not 100% happy with this, means having to add another route if the format is ever changed. Will continue to think.

Happy to kick this problem down the road. We don't need to solve it now.

thewilkybarkid:
Think this is all good now, just need to resolve/bypass the Hypertest problem.

thewilkybarkid:
https://github.com/hypermedia-app/hydra-validator/pull/57 should fix Hypertest.

thewilkybarkid (approved):

PR #194, opened 2020-01-21
Simplify creating HTTP errors

thewilkybarkid:
Contrary to its documentation, the `http-errors` module doesn't create a new error if it is already an HTTP error, so we don't need to check its type locally.


giorgiosironi:
I'd like to try to write a characterization test given it's an undocumented behavior if this is not urgent - more for my learning than for likelihood of future breakages which is minimal.

thewilkybarkid:
More meant that it doesn't replicate the same error, rather than being something that unexpected.

nlisgo (approved):
@thewilkybarkid is there a related open issue on the `http-errors` project. At least to address the documentation?

thewilkybarkid:
Commented at https://github.com/jshttp/http-errors/pull/37#issuecomment-576664583.

PR #195, opened 2020-01-21
Use http-status-codes library

thewilkybarkid:
Uses a dedicated library for HTTP status codes rather than Node's HTTP/2 module as it provides a simpler API.

### DIFF
--- a/test/app.test.ts
+++ b/test/app.test.ts
@@ -1,3 +1,4 @@
+import { NOT_FOUND, OK } from 'http-status-codes';
 import parseLinkHeader from 'parse-link-header';
 import request from 'supertest';
 import InMemoryArticles from '../src/adaptors/in-memory-articles';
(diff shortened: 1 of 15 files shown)


test/app.test.ts
```diff
@@ -20,13 +21,13 @@ describe('the application', (): void => {
...
 
-    expect(response.status).toEqual(200);
+    expect(response.status).toBe(OK);
   });
 
   it('should respond with 404 Not Found on an unknown path', async (): Promise<void> => {
     const response = await request(app.callback()).get('/does-not-exist');
 
-    expect(response.status).toEqual(404);
+    expect(response.status).toBe(NOT_FOUND);
```
thewilkybarkid, line 30: `toBe()` is used everywhere else, so changing here too.

giorgiosironi (approved):
Stable library and support seems complete

PR #196, opened 2020-01-22
Extract test middleware that always throws an error

thewilkybarkid:
Extracts the middleware used by the error handler test that always throws an error so it can be used elsewhere (https://github.com/libero/article-store/pull/184).

### DIFF
--- a/test/middleware.ts
+++ b/test/middleware.ts
@@ -1,9 +1,15 @@
+import { UnknownError } from 'http-errors';
 import { ExtendableContext } from 'koa';
 import { Middleware } from 'koa-compose';
 import { AppContext } from '../src/app';
 
 export type NextMiddleware<CustomT extends ExtendableContext = AppContext> =
   (context: CustomT) => Promise<void>;
+
+export const throwingNext = <CustomT extends ExtendableContext = AppContext>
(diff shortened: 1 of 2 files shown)


test/middleware.ts
```diff
@@ -1,10 +1,16 @@
+import { UnknownError } from 'http-errors';
 import { ExtendableContext } from 'koa';
 import { Middleware } from 'koa-compose';
 import { AppContext } from '../src/app';
 
 export type NextMiddleware<CustomT extends ExtendableContext = AppContext> =
   (context: CustomT) => Promise<void>;
 
+export const throwingNext = <CustomT extends ExtendableContext = AppContext>
+  (error: UnknownError = new Error()): NextMiddleware<CustomT> => (): never => {
```
thewilkybarkid, line 10: Changed return type to `never`, and removed `async` (seems to make no difference).

giorgiosironi: Given there's no `await` inside roughly I would expect `async` not to be required, but who knows what kind of type checking is going on

giorgiosironi (approved):

giorgiosironi:
Uh, conflicts block the build at the moment.

PR #207, opened 2020-01-27
Allow dev and test environments to diverge

nlisgo:
Moving to a [persistent storage](https://github.com/libero/article-store/pull/116) introduces the need to have a minimal setup to run unit tests. The `dev` and `prod` environments will need to, at least, have the database container as a dependency.

### DIFF
--- a/.docker/docker-compose.test.yml
+++ b/.docker/docker-compose.test.yml
@@ -5,12 +5,3 @@ services:
     build:
       target: test
     image: libero/article-store:${IMAGE_TAG:-master}-dev
-    volumes:
-      - ../.eslintignore:/app/.eslintignore
-      - ../.eslintrc.js:/app/.eslintrc.js
-      - ../jest.config.js:/app/jest.config.js
-      - ../src/:/app/src
-      - ../package.json:/app/package.json
-      - ../package-lock.json:/app/package-lock.json
-      - ../test/:/app/test
-      - ../tsconfig.json:/app/tsconfig.json
--- a/.docker/docker-compose.test.yml
+++ b/.docker/docker-compose.test.yml
@@ -1,0 +1,16 @@
+version: '3.4'
+
+services:
+  app:
+    build:
+      target: test
+    image: libero/article-store:${IMAGE_TAG:-master}-dev
+    volumes:
+      - ../.eslintignore:/app/.eslintignore
+      - ../.eslintrc.js:/app/.eslintrc.js
+      - ../jest.config.js:/app/jest.config.js
+      - ../src/:/app/src
+      - ../package.json:/app/package.json
+      - ../package-lock.json:/app/package-lock.json
+      - ../test/:/app/test
+      - ../tsconfig.json:/app/tsconfig.json
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 
 SHELL = /usr/bin/env bash
 
-ifneq (${TARGET}, prod)
+ifndef TARGET
 TARGET = dev
 endif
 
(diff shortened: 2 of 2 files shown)


.docker/docker-compose.test.yml
```diff
@@ -0,0 +1,16 @@
+version: '3.4'
+
+services:
+  app:
+    build:
+      target: test
+    image: libero/article-store:${IMAGE_TAG:-master}-dev
+    volumes:
```
giorgiosironi: when we paired, this introduced duplication between the `docker-compose.*.yml` files but I had no good ideas on how to remove it (only bad ones)

Makefile
```diff
@@ -68,7 +68,7 @@ fix: export TARGET = dev
 fix: ## Fix linting issues in the code
 	${DOCKER_COMPOSE} run --rm app npm run lint:fix
 
-test: export TARGET = dev
+test: export TARGET = test
```
giorgiosironi, line 71: hidden assumption that was made here: potential integration tests that use the database would be run with a separate `integration-test` Make target

giorgiosironi: Personally *in terms of naming targets* I would prefer something like:

- `unit-test`: in-memory tests
- `integration-test`: tests that use the database
- `test`: `unit-test` plus `integration-test` if the former passes

But this is unrelated, would only be introduced with integration tests.

.docker/docker-compose.test.yml
nlisgo: The only way around this would be to create more composer files so that we could share some things between environments but this may overcomplicate the Makefile.

thewilkybarkid: Not sure a `test` environment should have code mounted in it.

Makefile
```diff
@@ -3,7 +3,7 @@
 
 SHELL = /usr/bin/env bash
 
-ifneq (${TARGET}, prod)
+ifndef TARGET
 TARGET = dev
 endif
```
thewilkybarkid, line 8: Slight difference in that `TARGET=pord` probably breaks badly rather than defaulting to `dev`. Neither's ideal, better would be an error if it's not in the set list. Wonder if we can extract a change here to cater for that.

thewilkybarkid:
> The `dev` and `prod` environments will need to, at least, have the database container as a dependency.

This is referring to running (unit) tests, linting etc, where the `dev` `app` container doesn't need a database. Maybe there's something else we can do (could be as simple as not having the dependency, need to check what the implications are).

giorgiosironi:
> could be as simple as not having the dependency

As long as there is a separate `docker-compose up postgres -d` in the Makefile, the dependency is not needed. It does not provide a strong ordering as it only starts the containers rather than introducing any wait.

.docker/docker-compose.test.yml
nlisgo: > Not sure a `test` environment should have code mounted in it.

I will remove the volumes from test and we can reintroduce them if need arises.

.docker/docker-compose.test.yml
```diff
@@ -0,0 +1,7 @@
+version: '3.4'
+
+services:
+  app:
+    build:
+      target: test
+    image: libero/article-store:${IMAGE_TAG:-master}-dev
```
giorgiosironi, line 7: Not understanding, if the original point was to run the unit tests quickly (without starting the database), removing the volumes means you don't see the code changes when starting a new container; hence you need to rebuild the image every time you run the unit tests?

nlisgo:
After discussion with @giorgiosironi  and @thewilkybarkid, we have agreed to not introduce strong dependencies in the docker compose files but maintain them as separate setup steps in the Makefile. I will introduce these changes to https://github.com/libero/article-store/pull/116

PR #212, opened 2020-01-28
Update GitHub Actions with Dependabot

thewilkybarkid:
Adds GitHub Actions to the Dependabot, so pinned actions can be used. Also allows us to be notified of new major versions (eg [`actions/checkout@v2`](https://github.com/actions/checkout/releases/tag/v2.0.0)).

In draft as it's [not in the documentation](https://dependabot.com/docs/config-file/), but does pass the validator.


thewilkybarkid:
Ah, found https://github.com/dependabot/dependabot-core/issues/1461 suggests that it won't work yet.

.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v1.2.0
```
thewilkybarkid, line 20: Will extract these changes.

PR #214, opened 2020-01-28
Add data factory and dataset quad type

thewilkybarkid:
Recent changes in `@types/rdf-js` allow a quad type on data factories, this adds support (and for datasets too).


giorgiosironi (approved):
I guess this makes the return types of the factory stricter (there's some Quad substituted for the BaseQuad default here) which means fewer possibilities to assign the wrong thing e.g have a [BlankNode as a Predicate](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/rdf-js/index.d.ts#L154) would be forbidden. Everything seems to compile

PR #250, opened 2020-02-20
Make unhandled rejections fail

thewilkybarkid:
By default Node just emits a warning if there is an unhandled promise rejection. If something like https://github.com/libero/article-store/blob/003d90db13fc993861195c811ce9be46c22871a6/src/tools/initdb.ts#L5-L7 fails the process doesn't exit with an error code. This enables strict mode, so the error is thrown as expected.


thewilkybarkid:
Can be reproduced by removing https://github.com/libero/article-store/blob/003d90db13fc993861195c811ce9be46c22871a6/Makefile#L41 and running `make dev`.

giorgiosironi:
Reproduced the fix:
```
> ts-node ./src/tools/initdb

Error: getaddrinfo ENOTFOUND db
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (dns.js:64:26)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @ initdb:dev: `ts-node ./src/tools/initdb`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @ initdb:dev script.
```

giorgiosironi (approved):

PR #253, opened 2020-02-21
Use type imports

thewilkybarkid:
TypeScript 3.8 introduced type-only imports.

Not sure if this is all the cases, and not sure what should happen with ordering (keep the same order or separate them?).


src/app.ts
```diff
@@ -1,8 +1,8 @@
 import cors from '@koa/cors';
-import Router, { RouterContext } from '@koa/router';
+import type Router from '@koa/router';
```
thewilkybarkid, line 2: > If you’ve used Flow before, the syntax is fairly similar. One difference is that we’ve added a few restrictions to avoid code that might appear ambiguous.

This is a bit annoying.

tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "module": "commonjs",
     "outDir": "build",
     "strict": true,
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "importsNotUsedAsValues": "error"
```
thewilkybarkid, line 8: Better if the linter could fix these, refs https://github.com/typescript-eslint/typescript-eslint/issues/1436#issuecomment-589504369.

PR #254, opened 2020-02-21
Use ECMAScript Modules rather than CommonJS

thewilkybarkid:
As the target is Node only, ES modules can be used. This produces more concise, readable, and native, imports.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
     "target": "es2019",
-    "module": "commonjs",
+    "module": "esnext",
+    "moduleResolution": "node",
     "outDir": "build",
     "strict": true,
     "esModuleInterop": true


thewilkybarkid:
For interest, build of `app.ts` has changed from

thewilkybarkid:
For interest, built `app.js` has changed from:

```js
"use strict";
var __importDefault = (this && this.__importDefault) || function (mod) {
    return (mod && mod.__esModule) ? mod : { "default": mod };
};
Object.defineProperty(exports, "__esModule", { value: true });
const cors_1 = __importDefault(require("@koa/cors"));
const koa_1 = __importDefault(require("koa"));
const koa_logger_1 = __importDefault(require("koa-logger"));
const api_documentation_link_1 = __importDefault(require("./middleware/api-documentation-link"));
const data_factory_1 = __importDefault(require("./middleware/data-factory"));
const dataset_1 = __importDefault(require("./middleware/dataset"));
const empty_response_1 = __importDefault(require("./middleware/empty-response"));
const error_handler_1 = __importDefault(require("./middleware/error-handler"));
const jsonld_1 = __importDefault(require("./middleware/jsonld"));
const routing_1 = __importDefault(require("./middleware/routing"));
const namespaces_1 = __importDefault(require("./namespaces"));
```

to

```js
import cors from '@koa/cors';
import Koa from 'koa';
import logger from 'koa-logger';
import apiDocumentationLink from './middleware/api-documentation-link';
import setDataFactory from './middleware/data-factory';
import addDatasets from './middleware/dataset';
import emptyResponse from './middleware/empty-response';
import jsonld from './middleware/jsonld';
import namespaces from './namespaces';
```

Which is the same as https://github.com/libero/article-store/blob/9e9e5d74d284e959abf917f9326512aa1c5a48a2/src/app.ts#L1-L13 without the type-only imports.

tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
     "target": "es2019",
-    "module": "commonjs",
+    "module": "esnext",
```
thewilkybarkid, line 4: `es2020` works too...

thewilkybarkid: https://stackoverflow.com/a/59787575/77199 suggests that `target` should remain `es2019` but can add `"lib": ["es2020"]`.

thewilkybarkid:
Hmm, what should I be doing on a standalone script?

stephenwf:
if you add `export {}` I think it will treat it as a module.

thewilkybarkid:
Hmm, it isn't a module though... (ie it's not something you would `import`)

stephenwf:
Hmm, maybe it needs the `.mjs` to pick it up

thewilkybarkid:
This is where I get very frustrated with the JS world. 😅

Need to have a proper look at https://github.com/microsoft/TypeScript/issues/18442#issuecomment-581738714.

stephenwf:
Here's hoping [deno](https://github.com/denoland/deno) picks up traction!

thewilkybarkid:
Can't get anything to work. 😕

stephenwf:
@thewilkybarkid I think you've got 3 options. 

* Make it a module (even though it's not) and let Node pick it up as a ES module.
* Switch `import { a, b, c } from './blah'`; to `const {a, b, c} = await import('blah')` which can be run from both ES and CJS modules. 
* Add a package.json to the tools folder with `"type": "module"` and any other overrides you need for the file.

CJS modules are synchronous, so every `require()` [blocks while it loads](https://github.com/nodejs/node/blob/master/lib/internal/modules/cjs/loader.js#L1223), meaning they can be run without awaiting and always in the known order. Whereas an ESM module is always [asynchronously](https://github.com/nodejs/node/blob/master/lib/internal/modules/esm/loader.js#L165) loaded. So ESM modules can load other ESM modules at the top level (sort of mimicking top level await.) but CJS modules _need_ to await them, since they cannot be `await`-ed when imported by another module, since CJS assumes that you _will_ always get a module immediately. Since Node _needs_ to assign your top level script as either CJS or MJS it requires the extension or it needs to find a module – or be guided by one of the overrides. 

Hope that helps!

thewilkybarkid:
Going to close as I've never been able to get this working.

PR #260, opened 2020-02-28
Use namespace builder library

thewilkybarkid:
Adds `@tpluscode/rdf-ns-builders`, which adds extends the namespace builder with types for known properties. This will help avoid typos.

~I'll leave this in draft until the library is a little more stable, and the `@zazuko/rdf-vocabularies` dependency is removed (as it's _huge_ and not needed at runtime).~


src/routes/add-article.ts
```diff
@@ -24,8 +24,8 @@ export default (): AppMiddleware => (
       throw new createHttpError.BadRequest(`Article must have a blank node identifier (${termToString(id)} given)`);
     }
 
-    if (request.dataset.match(id, schema('name')).size === 0) {
-      throw new createHttpError.BadRequest(`Article must have at least one ${termToString(schema('name'))}`);
+    if (request.dataset.match(id, schema.name).size === 0) {
+      throw new createHttpError.BadRequest(`Article must have at least one ${termToString(schema.name)}`);
```
thewilkybarkid, line 28: As `name` is now a defined property on the builder, TypeScript now correctly recognises it as a `NamedNode`.

PHPStorm seems a bit confused (though doesn't complain):
Screenshot 2020-02-28 at 10 53 33

src/app.ts
```diff
@@ -44,7 +46,10 @@ export default (
   app.use(addDatasets());
   app.use(jsonld({
     '@language': 'en',
-    ...namespaces,
+    hydra: hydra().value,
+    owl: owl().value,
+    rdf: rdf().value,
+    schema: schema().value,
```
thewilkybarkid, line 52: @tpluscode Maybe we could add a `base` or `root` function to get the base IRI as a string?

tpluscode: I thought about this, and I'm not sure. Do you find yourself often in need of the base URI as string?

It would have to have a unique name, like `schema.__base`. Unless I duplicated the `prefixes` export and generated a default export

```ts
import prefixes from '@tpluscode/rdf-ns-builders'

app.use(jsonld({
  ...prefixes
}))
```

And neither is great IMO

thewilkybarkid: > Do you find yourself often in need of the base URI as string?

Only case I think, as this is outside RDF/JS.

You're right, neither of those is great...

thewilkybarkid: Opened https://github.com/zazuko/rdf-vocabularies/issues/55 to try and use a library rather than hard-coding them here.

PR #261, opened 2020-02-28
Expand the README

thewilkybarkid:
Adds details about the app, how to run it, how to develop, and badges.


README.md
```diff
@@ -1,31 +1,151 @@
...
+     -e "DATABASE_NAME=article-store" \
+     -e "DATABASE_USER=user" \
+     -p 8080:8080 \
+     liberoadmin/article-store:latest
+   ```
+
+4. Access the Article Store entry point:
+
+   ```shell
+   curl --verbose localhost:8080
```
thewilkybarkid: Worth showing how to run a [Hydra console](https://generic.hypermedia.app/) instead?

README.md
```diff
@@ -35,9 +155,104 @@ make dev
...
+
+### Linting
+
+The app is linted using [ESLint]. It can be run by executing:
+
+```shell
+make lint
+```
+
+Problems can be automatically fix, where possible, by executing:
```
NuclearRedeye:
```suggestion
Problems can be automatically fixed, where possible, by executing:
```

README.md
```diff
@@ -35,9 +155,104 @@ make dev
...
-- Ask a question on the [Libero Community Slack](https://libero.pub/join-slack).
-- Read the [code of conduct](https://libero.pub/code-of-conduct).
+- Report a bug or request a feature on [GitHub][new issue].
+- Ask a question on the [Libero Community Slack].
+- Read the [code of conduct].
+
+License
+-------
+
+This software is released under the [MIT license][license]. Copyright © 2019 [eLife Sciences Publications, Ltd][eLife].
```
NuclearRedeye: Should be 2020 I guess

README.md
```diff
@@ -1,31 +1,151 @@
...
 
-[![Build Status](https://github.com/libero/article-store/workflows/CI/badge.svg?branch=master)](https://github.com/libero/article-store/actions?query=branch%3Amaster+workflow%3ACI)
+[![Build status][Build badge]][Build]
+[![Open issues][Open issues badge]][Open issues]
+[![Docker pulls][Docker pulls badge]][Docker image]
+[![License][License badge]][License]
+[![Slack][Slack badge]][Libero Community Slack]
 
-Implementation of a [Libero API](https://libero.pub/api).
+This app provides a HTTP API for creating, maintaining and reading articles.
```
davidcmoulton:
```suggestion
This app provides an HTTP API for creating, maintaining and reading articles.
```

README.md
thewilkybarkid: Following @BlueReZZ's advice, it's when the repo was created (otherwise some commits aren't covered).

README.md
```diff
@@ -1,31 +1,151 @@
...
 
-Requirements
+An article is an [RDF graph], where the root node is a [`http://schema.org/Article`][schema:Article]. It can be
+represented in [JSON-LD] like:
```
davidcmoulton:
```suggestion
represented in [JSON-LD] as: 
```

README.md
```diff
@@ -35,9 +155,104 @@ make dev
 
 You can now access the entry point at <http://localhost:8080>, or view the console at <http://localhost:8081>.
```
davidcmoulton, line 213: I can see how `8080` relates to https://github.com/libero/article-store/pull/261/files#diff-04c6e90faac2675aa89e2176d2eec7d8R92, but is there a way of configuring a different port from `8081` if desired?

README.md
davidcmoulton: Should it be a range? If so 2019-2020.

README.md
thewilkybarkid: These actually refer to the local Docker Compose config: https://github.com/libero/article-store/blob/1bdf8a2a1069534ee0cb3fef96053c74e6debe38/.docker/docker-compose.yml#L14 and https://github.com/libero/article-store/blob/1bdf8a2a1069534ee0cb3fef96053c74e6debe38/.docker/docker-compose.yml#L21

Not customisable at the moment, but can be.

README.md
giorgiosironi: This is more complicated than people think

README.md
davidcmoulton: Might be worth making that more explicit? Whenever I see port numbers in a local dev context I always want to know if / how I can change them.

README.md
BlueReZZ: A range is also fine - some people prefer that as it shows it's not outdated, and gives you belt-and-braces to cover code that existed this year. Copyrights last for 70 years under common law and work on a  "whoever got there first" basis until you start the argument, so it's best to match the inception date of the repo.

My opinion appears to be corroborated by SO (confirmation bias FTW):
> The copyright notice on a work establishes a claim to copyright. The date on the notice establishes how far back the claim is made. This means if you update the date, you are no longer claiming the copyright for the original date and that means if somebody has copied the work in the meantime and they claim its theirs on the ground that their publishing the copy was before your claim, then it will be difficult to establish who is the originator of the work.

src: https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

README.md
```diff
@@ -1,31 +1,151 @@
...
+     -e "POSTGRES_DB=article-store" \
+     -e "POSTGRES_USER=user" \
+     postgres:11.5-alpine
+   ```
+
+2. Run the database creation with an ephemeral Article Store container:
+
+   ```shell
+   docker run --rm \
+     --link article-store-database \
```
giorgiosironi, line 85: is this easier than pasting here a Docker Compose configuration that uses published images (rather than starting from source)? Declarative is easier to teardown and to port to other contexts, whereas these containers have to be manually cleaned up.

thewilkybarkid: These use published images.

Agree that Compose is easier, happy to _also_ include it. (Feel it's useful to explain what's happening too though. And the database setup complicates it.)

thewilkybarkid: https://github.com/libero/article-store/pull/261/commits/b32c172c74dc2f30024e619baeaf4588c62b1876

README.md
```diff
@@ -1,31 +1,208 @@
...
+
+4. Access the Article Store entry point:
+
+   ```shell
+   curl --verbose localhost:8080
+   ```
+
+#### Using [Docker Compose]
+
+1. Create a file called `docker-compose.yml`:
```
davidcmoulton, line 112: Don't need instruction to create the file as it exists in the repo?

davidcmoulton:
(A few occurrences): the first line of output running `curl --verbose localhost:8080` is `Rebuilt URL to: localhost:8080/`

So the text should probably read

```curl --verbose localhost:8080/```

from the get go.

thewilkybarkid:
Maybe `curl --include http://localhost:8080/` to just show the full response too.

(Though I don't see the `Rebuilt URL`.)

README.md
thewilkybarkid: This is a little different in that it's for running it elsewhere (uses the published image). Maybe need some clearer wording around it.

davidcmoulton: Ah, I see. Yes, I think some clearer wording would be helpful here.

README.md
thewilkybarkid: Let's follow-up with a change to make them customisable.

README.md
thewilkybarkid: Was the comment actually related to having accidentally committed the file? 😳 (Refs https://github.com/libero/article-store/pull/261/commits/bbded6f9aaec1e4c6affd6427ecf6ce05fb26c4d.)

README.md
thewilkybarkid: Opened https://github.com/libero/publisher/issues/383.

BlueReZZ (approved):
Very succinct, good use of collapsible sections. I've not tried to follow the instructions but assuming they all run, then it looks good.

giorgiosironi (approved):

PR #226, opened 2020-02-03
Use airbnb-typescript/base rather than airbnb-base directly

thewilkybarkid:
Adds [eslint-config-airbnb-typescript](https://github.com/iamturns/eslint-config-airbnb-typescript) which overrides some of the `airbnb-base` plugin where it doesn't cater for TypeScript correctly.

Gets rounds the config problem in https://github.com/libero/article-store/pull/225.

PR #268, opened 2020-03-06
Remove indentation on multiline generics

thewilkybarkid:
Make sure that a generic which is split over multiple lines has the same indentation level.

Partially extracted from https://github.com/libero/article-store/pull/226 (triggered one of the two cases).


giorgiosironi (approved):
No preference either way

PR #270, opened 2020-03-09
Remove state from middleware

thewilkybarkid:
Simplifies the type of middleware returned where possible, to ease typing in test improvements.

The `Middleware` type in Koa adds state information to the type in Koa-Compose, but these middleware will have no knowledge of application state so can use the simpler form.

### DIFF
--- a/src/middleware/dataset.ts
+++ b/src/middleware/dataset.ts
@@ -1,6 +1,7 @@
 import {
-  DefaultContextExtends, DefaultStateExtends, ExtendableContext, Middleware, Next, Request, Response,
+  DefaultContextExtends, ExtendableContext, Next, Request, Response,
 } from 'koa';
+import { Middleware } from 'koa-compose';
 import {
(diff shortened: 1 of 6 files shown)


src/middleware/dataset.ts
```diff
@@ -14,8 +15,7 @@ export type DatasetContext<Context extends DefaultContextExtends = DefaultContex
   DataFactoryContext<Context & { request: WithDataset<Request, Q>; response: WithDataset<Response, Q> },
   ExtendedDataFactory<Q>>;
 
-export default (): Middleware<DefaultStateExtends, DataFactoryContext<ExtendableContext,
-ExtendedDataFactory<BaseQuad>>> => (
+export default (): Middleware<DataFactoryContext<ExtendableContext, ExtendedDataFactory<BaseQuad>>> => (
```
giorgiosironi, line 18: ah, this it not the same `Middleware` as before which explains the difference in the generic types

giorgiosironi (approved):
Remains a weird use of generics to me, but is a simplification in how many generic types are involved

PR #272, opened 2020-03-09
Move Hydra validator and Hypertest to Make

thewilkybarkid:
Allows the Hydra validator and Hypertest to be run locally, rather than just in CI.


.github/workflows/ci.yml
```diff
@@ -59,8 +59,8 @@ jobs:
           - make integration-test
           - TARGET=dev .github/smoke-test.sh
           - TARGET=prod .github/smoke-test.sh
-          - .github/analyse.sh
-          - .github/hypertest.sh
+          - TARGET=prod make api-validate
+          - TARGET=prod make api-test
         include:
           - target: prod
             test: TARGET=prod .github/smoke-test.sh
```
thewilkybarkid, line 66: Unrelated, but not sure why this duplicate the above?

.github/workflows/ci.yml
```diff
@@ -59,8 +59,8 @@ jobs:
           - make integration-test
           - TARGET=dev .github/smoke-test.sh
           - TARGET=prod .github/smoke-test.sh
-          - .github/analyse.sh
-          - .github/hypertest.sh
+          - TARGET=prod make api-validate
+          - TARGET=prod make api-test
```
thewilkybarkid, line 63: These didn't use `prod` before, but they probably should.

This doesn't work as need `matrix.target` set (probably refs https://github.com/libero/article-store/pull/272#discussion_r389700363).

.github/workflows/ci.yml
giorgiosironi: It is not a duplicate, but an override of that item; modifies `{{ matrix.target }}` in https://github.com/libero/article-store/pull/272/files#diff-e9f950f17198d3d5e3122a44230a09b9R76 to run on the production image rather than the default development one.

thewilkybarkid: Ah, is different to Travis. Adds config rather than new jobs (https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-configurations-in-a-matrix-build).

.github/analyse.sh
```diff
@@ -1,12 +0,0 @@
-#!/bin/bash
-set -e
-
-function finish() {
-  make logs stop
```
giorgiosironi, line 5: logs won't be printed anymore? Not a big loss

thewilkybarkid: Probably useful to keep for when it fails.

Makefile
```diff
@@ -93,6 +93,14 @@ integration-test: ## Run the integration tests
 	$(MAKE) start-db
 	${DOCKER_COMPOSE} run --rm app npm run test:integration; exit=$$?; ${DOCKER_COMPOSE} down; exit $$exit
 
+api-validate: ## Run the API analysis
+	$(MAKE) start wait-healthy
+	docker run --rm --network host hydrofoil/hydra-analyser:0.2.0 http://localhost:8080/; exit=$$?; ${DOCKER_COMPOSE} down; exit $$exit
+
+api-test: ## Run the API tests
```
giorgiosironi, line 107: API tests says that these tests exercise the system through the API, but nothing specific about how they take advantage of hypermedia and/or what parts are generated/discovered. Is this the fairest description?

Makefile
```diff
@@ -93,6 +93,14 @@ integration-test: ## Run the integration tests
...
+	$(MAKE) start wait-healthy
+	docker run --rm --init --network host --mount "type=bind,source=$(CURDIR)/test/hypertest/,destination=/tests" hydrofoil/hypertest:_0.4.1 --baseUri http://localhost:8080/; exit=$$?; $(DOCKER_COMPOSE) logs; ${DOCKER_COMPOSE} down; exit $$exit
```
giorgiosironi: `exit=$$?; $(DOCKER_COMPOSE) logs; ${DOCKER_COMPOSE} down; exit $$exit` seems error prone to duplicate and maintain. Can it be sticked into another Makefile variable like `DOCKER_COMPOSE`?

thewilkybarkid: Make sense. Already exists, so will extract the change.

thewilkybarkid: https://github.com/libero/article-store/pull/274

Will need a small change to add the logs back in.

Makefile
thewilkybarkid: Could call it system testing?

giorgiosironi: Your call, would add something a qualifier like `hypermedia-api-testing` if that's the main aspect.

giorgiosironi (approved):
Comments are being addressed

Makefile
thewilkybarkid: Will keep as is. The main part of the app is the API, and the API is hypermedia, so don't think it's necessary to include it here.

giorgiosironi: :+1:

PR #273, opened 2020-03-09
Don't run the Hydra console unless it's needed

thewilkybarkid:
Currently the Hydra console is run everywhere, even in cases like CI where it isn't needed. This limits it to just when running `make dev`.


.docker/docker-compose.console.yml
```diff
@@ -0,0 +1,14 @@
...
+    image: tpluscode/hypermedia-app:version-0.12.2
+    environment:
+      API_ENTRYPOINTS: '{ "http://localhost:8080/": "Article Store" }'
+      BASE_URL: http://localhost:8080/
+    ports:
+      - '8081:8000'
+    healthcheck:
+      test: wget --quiet --tries=1 --spider http://localhost:8000/ || exit 1
+      interval: 5s
+      timeout: 1s
```
giorgiosironi, line 14: I agree the `depends_on` wasn't useful

giorgiosironi (approved):

PR #274, opened 2020-03-09
Simplify handling of exit code

thewilkybarkid:
Capturing the exit code from a command and safely shutting down the containers is duplicated and a bit complicated. This extracts a variable.

Refs https://github.com/libero/article-store/pull/272#discussion_r389729422.


Makefile
```diff
@@ -12,6 +12,8 @@ endif
 
 DOCKER_COMPOSE = docker-compose --file .docker/docker-compose.yml --file .docker/docker-compose.${TARGET}.yml
 
+EXIT = exit=$$?; $(MAKE) stop; exit $$exit
```
thewilkybarkid: Name suggestions?

thewilkybarkid: `STOP`?

giorgiosironi: If this wasn't inserted in a command, the original exit code will be used. This behavior is preserved and an additional stop is inserted, so I think `STOP` captures the intention.

giorgiosironi (approved):

PR #280, opened 2020-03-13
Test propagation of errors when retrieving an article

thewilkybarkid:
Adds a test to cover a failure from the article repository.

Found in #167.


test/broken-object.ts
```diff
@@ -0,0 +1,12 @@
...
+export default <Type extends object>(error: UnknownError = 'error'): Type => (
+  new Proxy(
+    {} as Type,
+    {
+      get(): never {
+        throw error;
+      },
+    },
+  )
+);
```
thewilkybarkid: Not sure what to think of this.

giorgiosironi: Seems a standard Stub (with a failing behavior of course)

thewilkybarkid: Just wondering if there's a better way of doing it.

giorgiosironi: The interaction with `Articles` is explicitly calling a `get` method:
https://github.com/libero/article-store/blob/master/src/routes/article.ts#L15
`Proxy`, I discover today, is to add behavior on missing properties but there should be no (scalar) property access on `Articles`. Therefore, could drop the `Proxy` and just return an empty object with a `get` function on it throwing?

thewilkybarkid: The `get` on the `Proxy` is different to the `Articles.get` (refers to _any_ method rather than one named `get`). Have to fake the typing otherwise, so `{} as Article` (possibly even `{} as unknown as Article`), unless it's all implemented.

giorgiosironi: So it's a matter of generating something that fits the type? Statically typed languages often introduce [mocking frameworks](https://site.mockito.org/) for this reason.

thewilkybarkid: Maybe worth trying https://github.com/marchaos/jest-mock-extended.

thewilkybarkid: https://github.com/libero/article-store/pull/280/commits/4b7c803ea0157510890395ed6623726d30da5f99

stephenwf: Is there not a way that you could construct a real, but broken article?

PR #293, opened 2020-03-25
Wait for the Postgres container to be healthy

thewilkybarkid:
Some builds have been failing (eg https://github.com/libero/article-store/pull/282/checks?check_run_id=533861746), presumably due to the `initdb` script trying to be run before Postgres is ready. This adds an explicit wait for the container to be healthy before running the app.

Refs https://github.com/libero/article-store/pull/116#discussion_r372478291.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ start-db:
 	${DOCKER_COMPOSE} up --detach db
 
 init-db:
-	$(MAKE) start-db
+	$(MAKE) start-db wait-healthy
 	${DOCKER_COMPOSE} run --rm app npm run ${INITDB}
 
 stop: ## Stop the containers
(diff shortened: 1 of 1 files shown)


giorgiosironi:
I don't know how the driver is able to say:
```
error: the database system is starting up
```
but it reports a very specific (and language independent) code, `57P03`.

giorgiosironi:
If this is truly returned by Postgresql rather than its host being unreachable, that invalidates https://github.com/libero/article-store/pull/116#discussion_r372478291 rather than confirming it. That said, [pg_isready](https://www.postgresql.org/docs/11/app-pg-isready.html) which is [used in the healthcheck](https://github.com/libero/article-store/blob/master/.docker/docker-compose.yml#L24) is a stronger constraint that actually tries to successfully connect.

giorgiosironi (approved):

thewilkybarkid:
Interestingly, builds seem to now be consistently failing.

PR #318, opened 2020-04-06
Remove pg dependency

thewilkybarkid:
`pg-promise` has its own version constraint on `pg`, separate to the app. To avoid a mismatch, this removes `pg` in favour of using the typings of `pg-promise`.

Refs https://github.com/libero/article-store/pull/116/commits/68693d509430ec4f6e29b5385314eaf604be4227 and  https://github.com/libero/article-store/pull/308#issuecomment-609657109.


giorgiosironi (approved):
(assuming the `pg` module is completely hidden now)

